PR #4, opened 2016-11-28
CardanoCSL: fetch node0 public IP

domenkozar:
cc @georgeee

### DIFF
--- a/CardanoLib.hs
+++ b/CardanoLib.hs
@@ -1,9 +1,19 @@
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE OverloadedStrings #-}
+
 module CardanoLib where
 
 import Turtle
+import Data.Char (ord)
 import Data.Monoid ((<>))
 import qualified Data.Text as T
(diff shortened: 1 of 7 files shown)


CardanoLib.hs
```diff
@@ -69,7 +79,50 @@ scpFromNode args node from to = do
 
 sshForEach args cmd = nixops <> "ssh-for-each" <> args <> " -- " <> cmd
 
+-- Functions for extracting information out of nixops info command
+
 -- | Get all node IPs in EC2 cluster
+getNodes :: NixOpsArgs -> IO [Text]
```
georgeee: Please filter out 'obsolete' nodes

domenkozar: When is a node obsolete?

georgeee: When it's deployed and then commented out from nic configuration without destroying. Sometimes it's useful. This nodes are marked in `nixops info` output. Please try commenting out some nodes and check `info` output

domenkozar: Thanks! I'll check.

PR #6, opened 2016-12-12
[CSL-284] Implement enableP2P

domenkozar:


georgeee:
>This means also node0 will get peer flags
Fine by me :)

PR #9, opened 2017-01-10
Csl 517

domenkozar:
Using:

```
+  (genAttrs' (range 1 2) (key: "node${toString key}") (name: cardano-node-eu name)) // 
+  (genAttrs' (range 3 4) (key: "node${toString key}") (name: cardano-node-us name)) // 
+  (genAttrs' (range 5 6) (key: "node${toString key}") (name: cardano-node-asia name)) //
```

```
$ ./CardanoCSL.hs -c csl-staging.yaml firewall-block-region -f eu-central-1 -t ap-southeast-1
Blocking nodes: node5, node6
Applying on nodes: node0, node1, node2
```

cc @georgeee

Question: should node0 also be part of blocking logic?

### DIFF
--- a/CardanoCSL.hs
+++ b/CardanoCSL.hs
@@ -42,6 +42,8 @@ data Command =
   | DumpLogs { withProf :: Bool }
   | Start
   | Stop
+  | FirewallBlock { fromRegion :: Text, toRegion :: Text }
+  | FirewallClear
   | AMI
   | PrintDate
   deriving (Show)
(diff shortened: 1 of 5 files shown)


georgeee:
> Question: should node0 also be part of blocking logic?
Yes, according to it's region

georgeee:
Also @domenkozar, can we add command to reset all firewalls? I mean reset all changes made with `firewall-block-region` for all nodes. This would be useful to start experiment

georgeee:
Forget, hah. I only switched to code to see it's done

PR #10, opened 2017-03-13
Refactor Nix package generation

domenkozar:
Previously we were using `haskellPackageGen`, which generates packages
from their source, but this is using import-from-derivation
feature in Nix, meaning that evaluation and realization mix and match.

It was especially problematic for Hydra as it was building packages during evaluation and it timed out after 1h.

Now we have a `pkgs/generate.sh` file that also doesn't need to specify
sha256 so bumping packages is easier.

Soon that should be all automated using `stack2nix`.

### DIFF
--- a/modules/cardano-node.nix
+++ b/modules/cardano-node.nix
@@ -7,7 +7,7 @@ let
   node0 = nodes.node0.config;
   name = "cardano-node";
   stateDir = "/var/lib/cardano-node/";
-  cardano = (import ./../srk-nixpkgs.nix { inherit pkgs; inherit (cfg) genesisN slotDuration networkDiameter mpcRelayInterval; }).hspkgs.cardano-sl;
+  cardano = (import ./../srk-nixpkgs.nix { inherit pkgs; }).hspkgs.cardano-sl-static;
   distributionParam = "(${toString cfg.genesisN},${toString cfg.totalMoneyAmount})";
   enableIf = cond: flag: if cond then flag else "";
   smartGenIP = builtins.getEnv "SMART_GEN_IP";
--- a/srk-nixpkgs.nix
+++ b/srk-nixpkgs.nix
@@ -1,251 +1,54 @@
 { pkgs ? (import <nixpkgs> {})
 , compiler ? pkgs.haskell.packages.ghc802
-, genesisN ? 3
-, slotDuration ? 20
-, networkDiameter ? 6
-, mpcRelayInterval ? 16 } :
+}:
 
 with (import <nixpkgs/pkgs/development/haskell-modules/lib.nix> { inherit pkgs;});
 
 let
-  overrideAttrs = package: newAttrs:
-    package.override (args: args // {
-      mkDerivation = expr: args.mkDerivation (expr // newAttrs);
-    });
-
-  # Autogenerate default.nix from cabal file in src
-  haskellPackageGen = { doHaddock ? false, doFilter ? false, doCheck ? true, profiling ? false }: src:
-     let filteredSrc = builtins.filterSource (n: t: t != "unknown") src;
-         package = pkgs.runCommand "default.nix" {} ''
-           ${compiler.cabal2nix}/bin/cabal2nix \
-             ${if doFilter then filteredSrc else src} \
-             ${if doHaddock
-                 then ""
-                 else "--no-haddock"} \
-             ${if doCheck
-                 then ""
-                 else "--no-check"} \
-             ${if profiling
-                 then "--enable-profiling"
-                 else ""} \
-             > $out
-         '';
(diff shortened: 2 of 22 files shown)


domenkozar:
This is ready and uses latest cardano-sl and explorer.

cc @georgeee @volhovM @neongreen

domenkozar:
Before we deploy 0.3.0, I'd like to merge this and use newer nixpkgs commit and regenerate the ami.

srk-nixpkgs.nix
```diff
@@ -1,189 +1,54 @@
...
+      cardano-report-server = super.callPackage ./pkgs/cardano-report-server.nix { };
+      cardano-crypto = super.callPackage ./pkgs/cardano-crypto.nix { };
+      plutus-prototype = super.callPackage ./pkgs/plutus-prototype.nix { };
+      network-transport = super.callPackage ./pkgs/network-transport.nix { };
+      network-transport-tcp = super.callPackage ./pkgs/network-transport-tcp.nix { };
+      derive = super.callPackage ./pkgs/derive.nix { };
+      cryptonite = super.callPackage ./pkgs/cryptonite.nix { };
+
+      # TODO: https://github.com/NixOS/cabal2nix/issues/261
+      cardano-sl-core = prodMode (super.callCabal2nix "cardano-sl-core" "${self.cardano-sl.src}/core" {});
```
volhovM: How did you check that only `core` and `explorer` need to have `prodMode` option?

srk-nixpkgs.nix
```diff
@@ -1,189 +1,54 @@
 { pkgs ? (import <nixpkgs> {})
 , compiler ? pkgs.haskell.packages.ghc802
-, genesisN ? 3 
-, slotDuration ? 20 
-, networkDiameter ? 6
-, mpcRelayInterval ? 16 } :
```
volhovM: Aren't we planning to do more benchmarking? Or were these options moved to CLI?

modules/cardano-node.nix
```diff
@@ -7,7 +7,7 @@ let
   node0 = nodes.node0.config;
   name = "cardano-node";
   stateDir = "/var/lib/cardano-node/";
-  cardano = (import ./../srk-nixpkgs.nix { inherit pkgs; inherit (cfg) genesisN slotDuration networkDiameter mpcRelayInterval; }).hspkgs.cardano-sl;
+  cardano = (import ./../srk-nixpkgs.nix { inherit pkgs; inherit (cfg) genesisN slotDuration networkDiameter mpcRelayInterval; }).hspkgs.cardano-sl-static;
```
volhovM: You inherit these values here, but they are not used in `srk-nixpkgs`.

srk-nixpkgs.nix
domenkozar: This is long time not used anymore. One can now just fork cardano-sl and change the yaml file. It's not much different than changing it from Nix.

I don't know how Phillip is doing it now, but this doesn't work.

srk-nixpkgs.nix
domenkozar: I didn't check, just got it to compile :) should we set it for all cardano-* packages?

modules/cardano-node.nix
domenkozar: Thanks!

srk-nixpkgs.nix
volhovM: @neongreen will we have these constants in CLI after CLI refactoring?

srk-nixpkgs.nix
volhovM: It's used (defined in cabal file) in core, infra. Not used in db, lrc, update. So please add to infra as well, yes.

srk-nixpkgs.nix
neongreen: Is it fine if I make it an env var instead?

domenkozar: Why not just a yaml file like it's now? `--constants foo.yaml`

volhovM: I guess it's not urgent, just asked because thought you had been working with making our configs more extensible. Let's close this discussion for now. It's pretty easy to solve the problem if it arises in hour or so.

srk-nixpkgs.nix
domenkozar: Will push a fix now, thanks!

domenkozar:
@volhovM all concerns should be addressed :)

domenkozar:
@volhovM this now conflicts due to recent changes. I'll update to build `cardano-sl-0.3` and merge.

domenkozar:
Rebased on top of master, I'll merge this today not to get more conflicts.

PR #14, opened 2017-04-07
[DEVOPS-2] tarsnap: backup deployer

domenkozar:
# Changes

- set ami only for cardano nodes
- configure tarsnap backups for cardano-deployer
- bump nixops to be able to set `destDir` to avoid non-persistent key storage

# TODO

- [x] store secret persistently
- [x] create an issue for snapshots -> DEVOPS-120
- [x] update deployment plan to generate tarsnap key
- [x] depends on https://github.com/NixOS/nixops/pull/664

# Deploy

- [ ] Bump nixpkgs to fe62c993b5dfecb871a54eb6654b09bcd5595fe5
- [ ] create root key
- [ ] backup root key

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ txgen.json
 smart-gen-*
 static/id_buildfarm
 *prof
+static/*.secret
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ env:
 install:
   - nix-env -iA nixopsUnstable -f '<nixpkgs>'
   - touch datadog.secret
+  - echo "secret" > static/tarsnap-cardano-deployer.secret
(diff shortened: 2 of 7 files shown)


domenkozar:
I believe this shouldn't be merged yet, we should have a way to enable tarsnap per cluster, so it's only running for production. Same will be needed for other services (datadog, papertrail) so better get it done early.

domenkozar:
This needs to be rebased on top of #17 once it's merged, so we can only enable backups for production.

domenkozar:
By default nixops keys are deployed to `/run/keys` which is stored in-memory so if we reboot the machine backups will stop working.

I'd rather have no backups than not-working backups, so we need to fix this.

PR #17, opened 2017-04-12
[DEVOPS-3] Setup datadog for prod nodes and hydra/deployer

domenkozar:
TODO:

- [x] instructions how to include datadog into production nodes for next deploy
- [x] copy datadog api key to staging
- [x] test node cluster deploy
- [x] decide where is the single source of truth for changes in next deploy

### Notes for future deploys

- bump nixpkgs to 7ad99e9fc8c1977f4d51ba8531386fce7276d300
- add `deployments/cardano-prod.nix` to `deploymentFiles` section of yaml

### DIFF
--- a/deployments/cardano-prod.nix
+++ b/deployments/cardano-prod.nix
@@ -1,0 +1,12 @@
+with (import ./../lib.nix);
+
+let
+  conf = { config, pkgs, resources, ... }: {
+    imports = [
+      ./../modules/datadog.nix
+    ];
+    services.dd-agent.tags = ["prod"];
+  };
+in {
+  report-server = conf;
+} // (genAttrs' (range 0 13) (key: "node${toString key}") (name: conf))
--- a/modules/datadog.nix
+++ b/modules/datadog.nix
@@ -1,0 +1,13 @@
+{ config, pkgs, lib, ... }:
+
+with (import ./../lib.nix);
+
+{
+  config = {
+    services.dd-agent = {
+      enable = true;
+      api_key = fileContents ./../datadog.secret;
+      hostname = config.networking.hostName;
+    };
+  };
+}
--- a/deployments/infrastructure.nix
+++ b/deployments/infrastructure.nix
@@ -15,7 +15,11 @@ with (import ./../lib.nix);
       ./../modules/hydra-master.nix
       ./../modules/common.nix
       ./../modules/amazon-base.nix
+      ./../modules/datadog.nix
     ];
+
+    # TODO: move into separate -prod.nix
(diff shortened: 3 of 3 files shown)


deployments/cardano-prod.nix
```diff
@@ -0,0 +1,12 @@
...
+let
+  conf = { config, pkgs, resources, ... }: {
+    imports = [
+      ./../modules/datadog.nix
+    ];
+    services.dd-agent.tags = ["prod"];
+  };
+in {
+  report-server = conf;
+} // (genAttrs' (range 0 13) (key: "node${toString key}") (name: conf))
```
jmitchell: Can we feasibly determine the node names from the deployment configuration rather than hardcoding them?

It's also unclear how `node0`, e.g., is sufficiently descriptive. We have multiple distinct `node0`s in different clusters reachable from the production deployer.

modules/datadog.nix
```diff
@@ -0,0 +1,13 @@
+{ config, pkgs, lib, ... }:
+
+with (import ./../lib.nix);
+
+{
+  config = {
+    services.dd-agent = {
```
jmitchell: The name `dd-agent` comes from here, right? It would be nice to use `datadog-agent` to avoid confusion in the future.

domenkozar: The name comes from NixOS upstream: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/monitoring/dd-agent.nix

I agree abbreviations are confusing, we have a way to rename module attributes, but I'd leave this out of scope for this issue.

deployments/cardano-prod.nix
domenkozar: Good points!

There are two refactorings needed for this, and they've been sitting on the back of my mind for some time. I don't want to block this PR, as we need this soon, but we should make sure they come up quickly:

1)  make number of nodes a deployment input that can be specified in the config file and reused all over the place
2) separate EC2 specific settings into `explorer-prod.nix` and `explorer-staging.nix`, mostly based on `modules/amazon-base.nix` (this would allow us to build AMI without hacks like `generatingAMI`

I've opened DEVOPS-37 to address this, as they both touch the same code.

Concerning `node0` up to `nodeX`: these are name of the nodes in a cluster. Cluster is uniquely identified by deployment name, nodes always go from 0 to X. If you have a suggestion how to make this better (except from moving the number of nodes to a config file), I'm all ears :)

modules/datadog.nix
jmitchell: I agree this is out of scope. Thanks for clarifying.

PR #27, opened 2017-04-21
[DEVOPS-67] disable burst limiting of logs

domenkozar:
This also includes two fixes already deployed to testnet. Need to get them in ASAP for any further deploys.

### DIFF
--- a/deployments/cardano-prod.nix
+++ b/deployments/cardano-prod.nix
@@ -7,6 +7,9 @@ let
       ./../modules/papertrail.nix
     ];
     services.dd-agent.tags = ["prod"];
+
+    # DEVOPS-64: disable log bursting
+    services.journald.rateLimitBurst = 0;
   };
 in {
   report-server = conf;
--- a/deployments/cardano-prod.nix
+++ b/deployments/cardano-prod.nix
@@ -6,10 +6,13 @@ let
       ./../modules/datadog.nix
       ./../modules/papertrail.nix
     ];
-    services.dd-agent.tags = ["prod"];
+    services.dd-agent.tags = ["env:production"];
 
     # DEVOPS-64: disable log bursting
     services.journald.rateLimitBurst = 0;
+
+    # Initial block is big enough to hold 3 months of transactions
+    deployment.ec2.ebsInitialRootDiskSize = mkForce 700;
   };
 in {
   report-server = conf;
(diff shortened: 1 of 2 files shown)


deployments/cardano-prod.nix
```diff
@@ -6,7 +6,13 @@ let
       ./../modules/datadog.nix
       ./../modules/papertrail.nix
     ];
-    services.dd-agent.tags = ["prod"];
+    services.dd-agent.tags = ["env:production"];
+
+    # DEVOPS-64: disable log bursting
+    services.journald.rateLimitBurst = 0;
```
deepfire: Hmm, AIUI, without this option, `journald` would trim an excessively chatty stream of similar messages to 1 summary message per 30s.

This change effectively disables that trimming, forcing all messages to appear in the log.

How would that help?

deployments/cardano.nix
```diff
@@ -86,8 +84,13 @@ in
...
       # TODO: generalize with node generation
-      (genAttrs' (range 0 5) (key: "nodeip${toString key}") (name: { region = "eu-central-1"; inherit accessKeyId; })) //
-      #(genAttrs' (range 7 13) (key: "nodeip${toString key}") (name: { region = "us-west-1"; inherit accessKeyId; })) //
+      (genAttrs' (range 0 1) (key: "nodeip${toString key}") (name: { region = "eu-central-1"; inherit accessKeyId; })) //
+      (genAttrs' (range 2 3) (key: "nodeip${toString key}") (name: { region = "eu-west-1"; inherit accessKeyId; })) //
+      (genAttrs' (range 4 5) (key: "nodeip${toString key}") (name: { region = "eu-west-2"; inherit accessKeyId; })) //
+      (genAttrs' (range 6 7) (key: "nodeip${toString key}") (name: { region = "ap-southeast-1"; inherit accessKeyId; })) //
+      (genAttrs' (range 8 9) (key: "nodeip${toString key}") (name: { region = "ap-southeast-2"; inherit accessKeyId; })) //
+      (genAttrs' (range 10 11) (key: "nodeip${toString key}") (name: { region = "ap-northeast-1"; inherit accessKeyId; })) //
+      (genAttrs' (range 12 13) (key: "nodeip${toString key}") (name: { region = "ap-northeast-2"; inherit accessKeyId; })) //
```
deepfire: This effectively makes `cardano.nix` to correspond to the production version, right?

domenkozar: Correct, as it should be.

deployments/cardano-prod.nix
domenkozar: Currently, logs are trimmed, so we could be missing important information in log events. This helps us collect all logs.

PR #36, opened 2017-04-28
Added deployment script for explorer

ksaric:
Call the script with `deploy-explorer-linux.sh 1b80da3d18ab8903ec8577f867edfa86524d7b7b` to deploy hash `1b80da3d18ab8903ec8577f867edfa86524d7b7b` on staging server.

### DIFF
--- a/scripts/deploy-explorer-linux.sh
+++ b/scripts/deploy-explorer-linux.sh
@@ -6,15 +6,15 @@ if [ -z "$HASH_COMMIT" ]
     exit 1
 fi
 
-cd ~/devops-86-deploy-staging-explorer/pkgs
+cd ../pkgs
 # MAC users, beware! Not sure if the flags are correct.
 sed -ri "s/cabal2nix https:\/\/github\.com\/input-output-hk\/cardano-sl-explorer\.git --no-check --revision(.*?)> cardano-sl-explorer.nix/cabal2nix https:\/\/github\.com\/input-output-hk\/cardano-sl-explorer\.git --no-check --revision $HASH_COMMIT > cardano-sl-explorer.nix/g" generate.sh
 ./generate.sh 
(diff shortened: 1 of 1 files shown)


scripts/deploy-explorer-linux.sh
```diff
@@ -0,0 +1,15 @@
+HASH_COMMIT=$1
+
+ssh staging@35.156.156.28
```
domenkozar: I'd rather see this being invoked within iohk-nixops repo for two reasons:

- not to expose IP of deployer box
- this will be used for different features as well so folder path won't match

scripts/deploy-explorer-linux.sh
```diff
@@ -0,0 +1,15 @@
+HASH_COMMIT=$1
```
domenkozar: Check if $1 is set first.

scripts/deploy-explorer-linux.sh
ksaric: You are right. Will change.

scripts/deploy-explorer-linux.sh
ksaric: Agreed.

domenkozar:
@ksaric I'd still remove `~/devops-86-deploy-staging-explorer/` reference and cd to relative locations so this script can be used from top-level checkout of `iohk-nixops`.

ksaric:
@domenkozar I just saw this. Sorry, it's a pain to find what's actually relevant (this) from pointless updates from Github. What did you have in mind? Not sure what you mean.

domenkozar:
Thanks! I'll revisit this after https://github.com/input-output-hk/iohk-nixops/pull/47 is done as it will be simpler for all of us :)

domenkozar:
This will be automatic as part of DEVOPS-22 when doing `iohk deploy`. Thanks!

PR #37, opened 2017-04-28
[DEVOPS-45] Monitor cardano-node process on datadog

domenkozar:
This requires upstream changes in nixpkgs: https://github.com/NixOS/nixpkgs/pull/25288

A single node using process monitoring is running at `[staging@cardano-deployer:~/devops-45-monitor-cardano-node-process]$ nixops ssh -d devops-41 node0`

### DIFF
--- a/deployments/cardano-prod.nix
+++ b/deployments/cardano-prod.nix
@@ -6,13 +6,24 @@ let
       ./../modules/datadog.nix
       ./../modules/papertrail.nix
     ];
-    services.dd-agent.tags = ["env:production"];
 
     # DEVOPS-64: disable log bursting
     services.journald.rateLimitBurst = 0;
 
     # Initial block is big enough to hold 3 months of transactions
     deployment.ec2.ebsInitialRootDiskSize = mkForce 700;
+
+    services.dd-agent.tags = ["env:production"];
+    services.dd-agent.processConfig = ''
+      init_config:
+
+      instances:
+      - name: cardano-node
+        search_string: ['cardano-node']
+        exact_match: False
+        thresholds:
+          critical: [1, 1]
+    '';
   };
 in {
   report-server = conf;


deployments/cardano-prod.nix
```diff
@@ -6,13 +6,24 @@ let
...
+    services.dd-agent.processConfig = ''
+      init_config:
+
+      instances:
+      - name: cardano-node
+        search_string: ['cardano-node']
+        exact_match: False
+        thresholds:
+          critical: [1, 2]
+    '';
```
deepfire: One reason why I find this difficult to review, is because the version of `nixpkgs` employed isn't immediately clear:
```
deepfire@andromedae:~/src/iohk-nixops$ ag 'git.*nixpkgs' | cat
Infra.hs:3:#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/464c79ea9f929d1237dbc2df878eedad91767a72.tar.gz
TimeWarp.hs:3:#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/464c79ea9f929d1237dbc2df878eedad91767a72.tar.gz
jobsets/default.nix:17:        value = "https://github.com/NixOS/nixpkgs.git ebbababd8f9cb49d039f11d58e4e49d8e02d7533";
jobsets/spec.json:14:         "nixpkgs": { "type": "git", "value": "https://github.com/NixOS/nixpkgs-channels.git nixos-16.09", "emailresponsible": false }
scripts/generate-explorer-frontend.sh:4:export NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/763e21e982370f67c126f92a1113ea949db3b6e0.tar.gz
pkgs/generate.sh:3:#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/464c79ea9f929d1237dbc2df878eedad91767a72.tar.gz
CardanoCSL.hs:3:#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/464c79ea9f929d1237dbc2df878eedad91767a72.tar.gz
```

I.e. there are three versions of nixops mentioned in different contexts.  Granted -- the most obvious candidate is `464c79ea9f929d1237dbc2df878eedad91767a72`.

deepfire: Neither `464c79ea9f929d1237dbc2df878eedad91767a72`, `ebbababd8f9cb49d039f11d58e4e49d8e02d7533`, nor `763e21e982370f67c126f92a1113ea949db3b6e0` do not appear to have `dd-agent.processConfig` either.. 

..and now, I see that `processConfig` is added by you, very recently, in `0bc033117f190bafc139995b0c57f618bb5ddf27`.

But at least I see now what `processConfig` is supposed to do : -)

deepfire: In short -- perhaps, given our constraints, this is a bit too much procedure to follow, but it seems that it would really help to use a single, centrally-specified, published version of `nixpkgs` across all of IOHK devops.

Maybe this is a tall order -- maybe this even has been considered and rejected.. feel free to shoot me down!

domenkozar: Yes, this is confusing.

We will have to unify nixpkgs commit across projects as that will assure better reproducibility. Moreover, our Hydra should build against one nixpkgs commit (or at least as few as possible).

Unfortunately, this isn't as easy as moving things around, as there are many ways to pin nixpkgs (in shell, or in Nix, or in `config.yaml`, etc).

At the very least we should track where it's all defined so we can make reason and simplify.

In this very PR I have linked the nixpkgs pull request that this change was tested with as datadog changes aren't merged upstream yet.

domenkozar: FYI: this is tracked in DEVOPS-32

deployments/cardano-prod.nix
```diff
@@ -6,13 +6,24 @@ let
...
+    services.dd-agent.tags = ["env:production"];
+    services.dd-agent.processConfig = ''
+      init_config:
+
+      instances:
+      - name: cardano-node
+        search_string: ['cardano-node']
+        exact_match: False
+        thresholds:
+          critical: [1, 2]
```
jmitchell: This is an [inclusive range](https://github.com/DataDog/integrations-core/blob/fe6d775dbac8ba16f209b3a8abef08d9e464240b/process/check.py#L413-L414) and therefore allows 2 `cardano-node`. Do we ever expect multiple `cardano-node` processes per host?

deployments/cardano-prod.nix
jmitchell: > Unfortunately, this isn't as easy as moving things around, as there are many ways to pin nixpkgs (in shell, or in Nix, or in config.yaml, etc).

We ought to standardize on one strategy.

> I have linked the nixpkgs pull request that this change was tested with as datadog changes aren't merged upstream yet.

When we're waiting on upstream changes this makes a lot of sense. However, we have to be careful to avoid stomping on each other. In the absence of a single location (per iohk-nixops branch) for setting the NIX_PATH, when there are multiple pending upstream changes we risk accidentally introducing regression.

Would it help to make an official IOHK fork of nixpkgs and coordinate changes there? Ultimately we benefit from maintaining parity with nixos/nixpkgs, so I'm not suggesting we diverge any more than necessary.

deployments/cardano-prod.nix
domenkozar: We were planning to use two nodes for DDOS protection, but specification changed so we can set this back to 1, thanks!

deployments/cardano-prod.nix
domenkozar: > We ought to standardize on one strategy.

Moving to DEVOPS-32

> When we're waiting on upstream changes this makes a lot of sense. However, we have to be careful to avoid stomping on each other. In the absence of a single location (per iohk-nixops branch) for setting the NIX_PATH, when there are multiple pending upstream changes we risk accidentally introducing regression.

As long as we wait for upstream and bump nixpkgs at correct locations, it shouldn't be an issue (once we have it documented ofc).

> Would it help to make an official IOHK fork of nixpkgs and coordinate changes there? Ultimately we benefit from maintaining parity with nixos/nixpkgs, so I'm not suggesting we diverge any more than necessary.

I'd like to avoid this if possible. Upstream moves fast and I'm among top 10 contributors, so I can provide valuable feedback on PRs from within IOHK. If we are in very fast hurry we can just pin a fork branch.

I'd rather wait a few days and make sure upstream gets feedback than rely on a forked branch that will do the job.

This might change in the future if it becomes unbearable, but I don't expect much trouble.

domenkozar:
Changes are now upstream, I'm going to make another test.

domenkozar:
Tested and it works. I'll merge and put this on schedule for deployment.

PR #41, opened 2017-05-04
Allow intero to be used to develop the Haskell scripts

deepfire:
This allows `intero` to be used for the `iohk-nixops` script development.

Note that this requires a matching version of the `intero` emacs package.

(Just as a side note -- that's how I install emacs with packages: https://github.com/deepfire/bin/blob/master/nix/emacs.nix)


deepfire:
@jmitchell, renamed to `nixpkgs-shell.json`, I think that should do that..

domenkozar:
It's not entirely clear to me why we need `stack.yaml` and `shell.nix`?

I'm not against intero support, but I think we should rather first make our nixops wrapper a proper haskell package and then add editor support.

domenkozar:
Also this adds yet another source of `nixpkgs` pinning. It seems in general it's too early to merge, as we haven't figured out the foundamentals here.

deepfire:
@domenkozar, `stack.yaml` is required for intero, unfortunately, `shell.nix` for establishing the dependencies for the haskell scripts.

deepfire:
@domenkozar, it seems the merge should be undone -- I've let myself to be convinced by Jake that we can incrementally improve upon this PR : -)

PR #44, opened 2017-05-11
[DEVOPS-93] Add staging cardano-sl builds

domenkozar:


jobsets/default.nix
```diff
@@ -29,6 +29,20 @@ let
       path = "jobsets/cardano.nix";
       description = "Cardano SL";
     };
+    cardano-sl-staging = {
```
deepfire: This is essentially same as the `cardano-sl`, but with `nixpkgs` and `jobsets` bumped, right, @domenkozar?

domenkozar: Correct :)

PR #49, opened 2017-05-20
[DEVOPS-121] Cardano sl faster build

deepfire:
Besides _partially_ addressing https://issues.serokell.io/issue/DEVOPS-36#comment=96-3716, this also sets up `NIX_PATH`, so that a number of tools needn't have `-I` specified -- and so this pulls towards unification of `nixpkgs` specification (don't remember the issue).

cc @domenkozar, @jmitchell 

TODO:

- [x] decide on testing being a part of all Hydra builds
- [x] decide on whether building all 13 executables is worth it for all Hydra builds

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,6 @@
 { pkgs ? (import <nixpkgs> {})
 , compiler ? pkgs.haskell.packages.ghc802
+, builder-threads ? 8
 }:
 
 with (import <nixpkgs/pkgs/development/haskell-modules/lib.nix> { inherit pkgs;});
@@ -11,13 +12,13 @@ let
   });
   socket-io-src = pkgs.fetchgit (removeAttrs (lib.importJSON ./pkgs/engine-io.json) ["date"]);
   cardano-sl-src = pkgs.fetchgit (removeAttrs (lib.importJSON ./pkgs/cardano-sl.json) ["date"]);
-  cores   = 4;
-  threads = 8;
+
+  builder-cores   = builtins.div builder-threads 2;
   buildSubset = subset: drv: overrideCabal drv (drv: {
     buildTarget = subset;
   });
   buildSpeedupFlagsMarlowScaling = drv : appendConfigureFlag drv
-    "--ghc-option=-rtsopts --ghc-option=+RTS --ghc-option=-N${toString threads} --ghc-option=-A128m --ghc-option=-n2m --ghc-option=-qb0 --ghc-option=-qn${toString cores} --ghc-option=-RTS";
+    "--ghc-option=-rtsopts --ghc-option=+RTS --ghc-option=-N${toString builder-threads} --ghc-option=-A128m --ghc-option=-n2m --ghc-option=-qb0 --ghc-option=-qn${toString builder-cores} --ghc-option=-RTS";
 in compiler.override {
   overrides = self: super: {
     # To generate these go to ./pkgs and run ./generate.sh
--- a/default.nix
+++ b/default.nix
@@ -72,7 +72,7 @@ in compiler.override {
     });
     cardano-sl-static =
       buildSpeedupFlagsMarlowScaling
-      (buildSubset ("cardano-node")
+      (# buildSubset ("cardano-node") -- unfortunately, we need customizing the .cabal file for this to work..
        (dontCheck (linkWithGold (justStaticExecutables self.cardano-sl))));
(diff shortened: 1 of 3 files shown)


default.nix
```diff
@@ -11,6 +11,13 @@ let
   });
   socket-io-src = pkgs.fetchgit (removeAttrs (lib.importJSON ./pkgs/engine-io.json) ["date"]);
   cardano-sl-src = pkgs.fetchgit (removeAttrs (lib.importJSON ./pkgs/cardano-sl.json) ["date"]);
+  cores   = 4;
```
domenkozar: These attributes should evaluate to a derivation. Instead threads and cores should be part of the function inputs for default.nix

default.nix
```diff
@@ -11,6 +11,13 @@ let
...
+  threads = 8;
+  buildSubset = subset: drv: overrideCabal drv (drv: {
+    buildTarget = subset;
+  });
+  buildSpeedupFlagsMarlowScaling = drv : appendConfigureFlag drv
+    "--ghc-option=-rtsopts --ghc-option=+RTS --ghc-option=-N${toString threads} --ghc-option=-A128m --ghc-option=-n2m --ghc-option=-qb0 --ghc-option=-qn${toString cores} --ghc-option=-RTS";
```
domenkozar: -N is already set from cabal in cardano-sl

domenkozar:
For `-static` we should skip tests and build only targets we need for deployments. `cardano-sl` should still be a full build with tests etc.

I'd like to see time differences here before/after this PR and with current VM and 16 core VM.

deepfire-pusher:
@domenkozar, the 16-core VM, should it be a new one, or the currently-provisioned `i3.2xlarge` is sufficient?

deepfire-pusher:
@domenkozar, regarding the second question from the PR TODO:

How do we subset the 13 executables -- are we clear on what is needed for deployments?

Should I judge this myself?

domenkozar:
So far we need `cardano-node` and `cardano-explorer` binaries (I just grepped `/bin/`)

domenkozar:
1) That should be good enough
2) 4 results: before/after, 2 core/8 core

deepfire-pusher:
@domenkozar, thanks, I'll try to do this today

domenkozar:
@deepfire-pusher I see now that you have updated the issue. Those statistics are good enough.

Let's build only `cardano-node` and `cardano-explorer` executables and that should be more than good enough.

domenkozar:
Note that nix already pases `-j$NIX_BUILD_CORES` to ghc, so it's not single threaded by default.

deepfire-pusher:
If it's already parallel.. well, I that's scary, it means that `hydra.iohk.io` is even slower than I thought : -)

domenkozar:
It's slower because there are multiple packages compiling together. Up to 4.

deepfire-pusher:
Ah, I see -- that was the problem I was trying to attack initially -- the direct comparability of measurements.

But I agree, just running the tests manually everywhere is faster, and the result is useful for judgement of effect of changes.

default.nix
deepfire: 1. The problem here is.. we need to set both `-N` and `-qn`, in the `2:1` ratio, as per Simon Marlow's suggestion (see https://ghc.haskell.org/trac/ghc/ticket/9221#comment:72).
2. Once we accept №1, a need to have an authoritative _Nix-evaluation-time_ source of information on the amount of cores on the system.
3. I couldn't find such a source, hence the hard-coding.
4. Since we hard-code `-qn`, I guess it's kinda debatable if leaving `-N` _free_ is a good thing.  It would certainly somewhat obscure the _Marlow ratio_ between cores and threads.

default.nix
deepfire: Agreed!

default.nix
deepfire: In fact, @domenkozar, I can't find where `cabal` sets `-N` -- at least it doesn't in https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/generic-builder.nix

Cabal _does_ set `-j`, but that's `ghc --make`-level parallelism.

deepfire:
@domenkozar, what is our next step here?

domenkozar:
This will be handled as part of DEVOPS-121.

domenkozar:
I'm going to close this, as expressions moved to cardano-sl repo and we can revisit the PR once DEVOPS-121 is picked up.

PR #51, opened 2017-05-24
[DEVOPS-53] enable firewall on nodes

deepfire:
Issue: https://issues.serokell.io/issue/DEVOPS-53

### DIFF
--- a/modules/cardano-explorer.nix
+++ b/modules/cardano-explorer.nix
@@ -14,6 +14,12 @@ with (import ./../lib.nix);
       autoStart = true;
     };
 
+    networking.firewall = {
+      allowedTCPPorts = [
+        80 # nginx
+      ];
+    };
+
     services.nginx = {
       enable = true;
       virtualHosts = {
--- a/modules/cardano-explorer.nix
+++ b/modules/cardano-explorer.nix
@@ -14,8 +14,6 @@ with (import ./../lib.nix);
       autoStart = true;
     };
 
-    networking.firewall.enable = false;
-
     services.nginx = {
       enable = true;
       virtualHosts = {
--- a/modules/cardano-explorer.nix
+++ b/modules/cardano-explorer.nix
@@ -14,9 +14,7 @@ with (import ./../lib.nix);
       autoStart = true;
     };
 
-    networking.firewall.allowedTCPPorts = [
-      80 # nginx
-    ];
+    networking.firewall.enable = false;
 
     services.nginx = {
       enable = true;
--- a/modules/cardano-node.nix
+++ b/modules/cardano-node.nix
@@ -142,7 +142,14 @@ in {
       "${toString hour} * * * * root /run/current-system/sw/bin/systemctl restart cardano-node"
     ];
 
-    networking.firewall.allowedTCPPorts = [ cfg.port ];
+    networking.firewall = {
+      enable = mkForce true;
+      allowedTCPPorts = [ cfg.port ];
+
+      # TODO: securing this depends on CSLA-27
+      # NOTE: this implicitly blocks DHCPCD, which uses port 68
+      allowedUDPPortRanges = [ { from = 1024; to = 65000; } ];
+    };
 
     systemd.services.cardano-node = {
       description   = "cardano node service";
--- a/modules/cardano-node.nix
+++ b/modules/cardano-node.nix
@@ -143,7 +143,6 @@ in {
     ];
 
     networking.firewall = {
-      enable = mkForce true;
       allowedTCPPorts = [ cfg.port ];
 
       # TODO: securing this depends on CSLA-27
--- a/modules/report-server.nix
+++ b/modules/report-server.nix
@@ -25,7 +25,7 @@ in {
       groups.report-server = {};
     };
 
-    networking.firewall.allowedTCPPorts = [ cfg.port ];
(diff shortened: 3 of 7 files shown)


deepfire:
Functionality of a 4-node cluster was tested with firewall enabled:
- nodes find each other and agree on the blockchain tip

deepfire:
Next step:  test cluster functionality with `dhcpcd` disabled.  I have my doubts that this will work, but am ready to be pleasantly surprised : -)

modules/cardano-node.nix
```diff
@@ -142,7 +142,14 @@ in {
       "${toString hour} * * * * root /run/current-system/sw/bin/systemctl restart cardano-node"
     ];
 
-    networking.firewall.allowedTCPPorts = [ cfg.port ];
+    networking.firewall = {
+      enable = mkForce true;
```
domenkozar: By default firewall is on. We should get rid of our `false` overrides in the codebase so that this line is redundant.

deepfire:
Disabling `dhcpcd` went smooth, despite my premonitions -- Domen was right. : -)

Next step is refactoring suggested by Domen.

deepfire:
@domenkozar, what would be a test plan for a refactoring with this kind of potential consequences?

domenkozar:
Consequence of my suggestion or this PR in general?

deepfire:
The refactoring you suggest, yes -- it would potentially affect the whole repository -- it shouldn't, of course, if implemented correctly.

The non-refactoring part has been tested, as I have described above.

domenkozar:
We'll have to test as part of next staging deploy that all services work. Really rest of the modules have always considered firewall as being on :)

Meaning, note in this issue that firewall has to be tested on explorer/reportserver as well.

deepfire:
@domenkozar, current situation on the `networking.firewall.enable` value:
- https://issues.serokell.io/_persistent/iohk-nixops-firewall-config.pdf?file=100-469&c=true

From this, I see the following:
- enabled:
   - `cardano-deployer`
   - `cardano-node`
   - `hydra`
- disabled:
   - `cardano-explorer`
   - `report-server`
   - `timewarp`

deepfire:
@domenkozar, pushed the refactoring of `networking.firewall.enable`.

modules/timewarp-node.nix
```diff
@@ -27,7 +27,7 @@ in {
       groups.timewarp = { };
     };
 
-    networking.firewall.allowedTCPPorts = [ cfg.port ];
```
domenkozar: Let's leave the port configured, even if firewall is off.

modules/report-server.nix
```diff
@@ -25,7 +25,7 @@ in {
       groups.report-server = {};
     };
 
-    networking.firewall.allowedTCPPorts = [ cfg.port ];
+    networking.firewall.enabled = false;
```
domenkozar: Actually for report-server, I'd just open up the port. I'm confident it will work.

modules/cardano-explorer.nix
```diff
@@ -14,9 +14,7 @@ with (import ./../lib.nix);
       autoStart = true;
     };
 
-    networking.firewall.allowedTCPPorts = [
```
domenkozar: We need firewall here enabled as well, let's revert this change. Explorer is also a cardano-node.

domenkozar: @deepfire removal of these port whitelisting should be reverted

modules/report-server.nix
```diff
@@ -25,8 +25,6 @@ in {
       groups.report-server = {};
     };
 
-    networking.firewall.allowedTCPPorts = [ cfg.port ];
```
domenkozar: @deepfire removal of these port whitelisting should be reverted

deepfire: @domenkozar, done

deepfire:
@domenkozar, there was  a problem testing this on staging: https://issues.serokell.io/issue/DEVOPS-53#comment=96-3955

I'm not really sure how to make it proper -- either we need to patch nginx (which.. well, doesn't seem all bad -- why would an option require the path to exist on nixops evaluator?), or we need to figure out how to make it not to be forced during nixops evaluation

deepfire:
Alright, after investigations, my conclusion was that we cannot disable `dhcpcd` on EC2 instances.  Kind of expected that (https://github.com/input-output-hk/iohk-nixops/pull/51#issuecomment-303846777) -- nodes don't come up.

The surprising part is how this slipped through my testing.

I have a theory -- `nixops modify` isn't as good as a `nixops destroy && nixops delete && nixops create` -- simply because the former doesn't restart the machines.

Hmm. makes sense again : -)

deepfire:
In any case, a freshly-spun 4-node cluster works with the tip of this branch -- and the blockchain tip is agreed upon -- see https://issues.serokell.io/issue/DEVOPS-53#comment=96-4012

@domenkozar, I think this is finally ready.

domenkozar:
@deepfire there are still 2 unaddressed comments

modules/timewarp-node.nix
deepfire: Done!

modules/cardano-explorer.nix
deepfire: Done

PR #54, opened 2017-06-01
[DEVOPS-58] Automate creating security groups

domenkozar:
This also changes `enableP2P` to be respected only in development since it doesn't make sense in production.

TODO:

- [x] this allows just all TCP, we should allow all traffic

### DIFF
--- a/deployments/keypairs.nix
+++ b/deployments/keypairs.nix
@@ -1,7 +1,39 @@
 with (import ./../lib.nix);
 
-{
+let
+  regions = [
+    "eu-central-1"
+    "eu-west-1"
+    "eu-west-2"
+    "ap-southeast-1"
+    "ap-southeast-2"
+    "ap-northeast-1"
+    "ap-northeast-2"
+  ];
+  genSG = sgs: mergeAttrs (map sgs regions);
+in {
   resources = {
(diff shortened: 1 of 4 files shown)


modules/cardano-node-aws.nix
```diff
@@ -4,16 +4,17 @@ testIndex: region:
   { config, resources, pkgs, nodes, ... }:
     let
       cfg = config.services.cardano-node;
+      cardanoNodes = filter (c: c.services.cardano-node.enable)
```
deepfire: I would have tried to find a more descriptive name for this,
like `enabledNodeConfigs`, or somesuch..

deployments/keypairs.nix
```diff
@@ -1,7 +1,39 @@
 with (import ./../lib.nix);
 
-{
+let
+  regions = [
```
deepfire: By the way, can we have this region list defined globally?

domenkozar: Yes, but it requires some refactoring. The good news is, if regions mismatch you'll get an error :)

modules/cardano-node-aws.nix
domenkozar: Fixed!

PR #55, opened 2017-06-02
[DEVOPS-123]  Provide NIX_PATH in nix-shell; add override tools

deepfire:
This pursues https://issues.serokell.io/issue/DEVOPS-132, by removing one point of `<nixpkgs>` variance:  environment.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -9,6 +9,13 @@ let
   prodMode = drv: overrideCabal drv (drv: {
     configureFlags = [ "-f-asserts" "-f-dev-mode"];
   });
+
+  githubSrc     =      repo: rev: sha256:       pkgs.fetchgit  { url = "https://github.com/" + repo; rev = rev; sha256 = sha256; };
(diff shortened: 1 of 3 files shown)


domenkozar:
I don't understand how this fixes the issue, can you elaborate?

deepfire:
@domenkozar, the environment has arbitrary value for `NIX_PATH`.

By issuing `nix-shell` inside `iohk-nixops` we enter a controlled environment with `<nixpkgs>` specified by `nixpkgs-src.json`.

As an example, this allows us to drop all the headers in the `*.hs` scripts, like:
- https://github.com/input-output-hk/iohk-nixops/blob/master/TimeWarp.hs#L3
- https://github.com/input-output-hk/iohk-nixops/blob/master/CardanoCSL.hs#L3
..etc.

Similarly, we can drop `-I nixpkgs=...` from many commands that we issue.

domenkozar:
Superseeded by https://github.com/input-output-hk/iohk-nixops/pull/77

PR #58, opened 2017-06-07
[DEVOPS-142]  Implement a global --debug flag for all deployment scripts

deepfire:
This is to:

1. allow `--debug` passed to most `nixops` commands
2. facilitate easier future options definitions

### DIFF
--- a/CardanoCSL.hs
+++ b/CardanoCSL.hs
@@ -71,14 +71,14 @@ subparser =
   <|> subcommand "ami" "Build ami" (pure AMI)
   <|> subcommand "generate-ipdht" "Generate IP/DHT mappings for wallet use" (pure GenerateIPDHTMappings)
 
-parser :: Parser (FilePath, Command)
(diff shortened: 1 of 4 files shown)


domenkozar:
Let's just always pass `--debug`. Not having tracebacks is a bad idea in the first place, because of issues like https://github.com/NixOS/nixops/issues/677

*nitpick: please always separate styling changes from features/bugs*

deepfire:
@domenkozar, I will undo the reindentation.

deepfire:
Ok, instead of undoing them, I separated them into a separate commit (the styling changes were fairly minor, though), @domenkozar.

deepfire:
@domenkozar, I have played with adding `--debug` to `nixops deploy` -- and the unconditional XML dump is absolutely _terrifying_.

I think we should leave `--debug` as an option.

What do you think?

domenkozar:
Makes sense, let's leave `--debug` as an option then.

PR #59, opened 2017-06-07
[DEVOPS-108] Datadog monitors in nix form

jmitchell:
The `appKey` still needs to be set, and the whole thing tested.

Before testing we need to ensure there are no problems syncing our infra deployments with the current master.


jmitchell:
Ready for review. After it's approval I will deploy and verify.

Are there concerns about doing a fresh infra deployment of master, or should I cherry-pick these changes on top of the latest deployment?

deployments/infrastructure-env-production.nix
```diff
@@ -37,5 +37,147 @@ with (import ./../lib.nix);
...
+        #
+
+        crit_cardano_node_process_monitor = { config, ... }:
+          {
+            inherit apiKey appKey;
+
+            name = "Critical: cardano-node process is down";
+            type = "metric alert";
+            message = messages.critical;
+            query = "\"process.up\".over(\"cardano-node\",\"process:cardano-node\").by(\"host\",\"process\").last(2).count_by_status()";
```
domenkozar: What's the interval over this will trigger? If we restart the nodes, will it go off? It should have trigger only after 5min of downtime.

deployments/infrastructure-env-production.nix
```diff
@@ -37,5 +37,147 @@ with (import ./../lib.nix);
...
+        #
+
+        crit_cpu_monitor = { config, ... }:
+          {
+            inherit apiKey appKey;
+
+            name = "Critical: high CPU usage";
+            type = "metric alert";
+            message = messages.critical;
+            query = "avg(last_5m):avg:system.load.norm.1{host:${config.deployment.name}.machine} by {host} > 0.9";
```
domenkozar: I'm thinking we should watch for 99th percentile instead of average to get rid of most of outliers.

domenkozar:
@jmitchell To test you can deploy fresh deployment (initially nodes). You have only added these monitors to infrastructure (hydra, deployer), but we also need them for cardano nodes, explorer and report server.

domenkozar:
2017-06-08-114151_604x339_scrot

This gives me thinking:

1) we will need different monitors per server (CPU on hydra will be really random, so we shouldn't even watch it)

2) it trigger low and high priorities? Is that due to the fact we will need two datadog accounts?

jmitchell:
> we will need different monitors per server

Agreed.

> it trigger low and high priorities? Is that due to the fact we will need two datadog accounts?

Yes, I believe so. As configured when the critical threshold is met the non-critical threshold for the separate non-critical monitor is also met, and both are triggered.

I just had an epiphany about how the notification mechanism works, and may be able to combine these into a single monitor with different notification paths based on urgency. I'll test it out and report in DEVOPS-108.

deployments/infrastructure-env-production.nix
jmitchell: The sampling/bucketing rate isn't completely clear to me even after scouring the docs. I've reached out to Datadog support to clarify how frequently service check monitors are sampled and translated into consecutive failures.

deployments/infrastructure-env-production.nix
jmitchell: A high percentile along with shorter sampling duration would be better. That doesn't seem to be offered as a default feature, though. We may have to [customize the the datadog agent configuration](https://help.datadoghq.com/hc/en-us/articles/211545826-Why-histogram-stats-are-all-the-same-inaccurate-Characteristics-of-Datadog-histograms-) for metrics like that.

In the short term for DEVOPS-108 we should focus on deploying a functional monitoring/alerting system, and we can fine tune as we go.

jmitchell:
Check out the latest comment in YT.

modules/datadog-monitors.nix
```diff
@@ -0,0 +1,69 @@
+with (import ./../lib.nix);
+
+rec {
+  mkMonitor = { name, type, query, message ? "", monitorOptions ? {} }:
```
domenkozar: Looks good :) I'd like to avoid this pattern of functions that return nixos modules, but rather use options in modules to enable monitors.

Something like `iohk.monitors = [ disk_monitor ntp_monitor ];`. @cleverca22 can help you with options, but we are using them for example in `modules/cardano-node.nix`

domenkozar:
FYI: while deploying staging environment, I got the following error:

```
error: ['Critical threshold (0.99) does not match that used in the query (0.9).']
```

I've changed it to `0.9` and it works.

jmitchell:
Ah, yes, I found that too and neglected to bring my ad-hoc fix back to this branch. There's a coupling between the `query` and the `monitorOptions.thresholds`. The diff should clarify:

```
diff --git a/deployments/cardano-nodes-env-staging.nix b/deployments/cardano-nodes-env-staging.nix
index b20ae44..01fc8c5 100644
--- a/deployments/cardano-nodes-env-staging.nix
+++ b/deployments/cardano-nodes-env-staging.nix
@@ -8,6 +8,7 @@ in {
     elasticIPs = mkNodeIPs nodes accessKeyId;
       cpu = mkMonitor (cpu_monitor // {
+        query = config: "avg(last_5m):avg:system.load.norm.1{host:${config.deployment.name}.machine} by {host} > 0.99";
         monitorOptions.thresholds = {
          warning = "0.95";
          critical = "0.99";
```

domenkozar:
Thanks!

PR #63, opened 2017-06-10
[DEVOPS-51]  Automation of `iohk-nixops` checkouts, AWS tooling and Route53 config

deepfire:
- extend `scripts/aws.hs` with route53 Hosted Zone tooling and a simple IAM `whoami` query
- provision `cardano-node-N.aws.iohkdev.io` DNS entries


scripts/aws.hs
```diff
@@ -1,17 +1,46 @@
-#!/usr/bin/env nix-shell
-#! nix-shell -j 4 -i runhaskell -p 'pkgs.haskellPackages.ghcWithPackages (hp: with hp; [ turtle ])'
-#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/f2c4af4e3bd9ecd4b1cf494270eae5fd3ca9e68c.tar.gz
```
domenkozar: This removes the ability to use the script directly, until we install it with Nix let's leave it around.

domenkozar:
Can you explain what `Route53` part in haskell does? I'd rather see we add any support to nixops than to maintin custom haskell

scripts/aws.hs
deepfire: I can do that, no problem, but..

What if we consider entering `nix-shell` a prerequisite for working with a `iohk-nixops` checkout?

We would get this:

1. ability to drop all these nixpkgs/package set lines from all these scripts.
2. a standartized environment
3. reduced script startup times -- since they are going to share the Nix cache for their dependencies -- since the `nixpkgs` is the same across all of them _and_ whatever builds of `cardano-sl`
4. a reduced zoo of different `nixpkgs` versions used

I really don't want to come across pushy, but I don't think we ever really discussed this in a satisfying manner.

deepfire: cc @domenkozar

deepfire:
As it stands, the `Route53` part in Haskell is not directly required for anything -- it was used to support experimentation.

It can be used to query the current contents of the `aws.iohkdev.io` DNS zone.

scripts/aws.hs
domenkozar: We can discuss this and I'm in favor of migrating to a proper packaging, but it's entirely out of scope for DEVOPS-51

domenkozar:
How exactly was `cardano-node-N.aws.iohkdev.io` provisioned (on cloudflare and route53)?

scripts/aws.hs
deepfire: Understood, will return the packages meanwhile : -)

deepfire:
1. Cloudflare delegates `aws.iohkdev.io` to the DNS servers that `Route53` provides for the `aws.iohkdev.io` _Hosted Zone_.
2. Since, as described in the DEVOPS-51 comments, `cardano-deployer` AWS user is unable to access _Route53_, the Hosted Zone RRS entries were manually inserted using `scripts/aws.hs`.

This is all described in https://issues.serokell.io/issue/DEVOPS-51#comment=96-4299

scripts/aws.hs
deepfire: Done.

modules/cardano-node-prod.nix
```diff
@@ -14,6 +14,9 @@ with (import ./../lib.nix);
 
   deployment.ec2.elasticIPv4 = resources.elasticIPs.${"nodeip" + toString config.services.cardano-node.testIndex};
 
+  deployment.route53.accessKeyId = resources.ec2KeyPairs.${keypairFor region}.accessKeyId;
+  deployment.route53.hostName = "cardano-node-${toString config.services.cardano-node.testIndex}.aws.iohkdev.io";
```
domenkozar: `iohkdev.io` should be used for `cardano-node-staging.nix`, `iohk.io` for `cardano-node-prod.nix`.

scripts/aws.hs
```diff
@@ -40,26 +69,67 @@ data RSpecWin64   = Win64 AppveyorPath  deriving (Show)
...
+withAWS ∷ Options → (forall r m . AWSConstraint r m ⇒ m a) → IO a
+withAWS Options{..} awsAction = do
+  lgr ← newLogger (if oDebug then Debug else Info) Sys.stdout
+  env ← newEnv Discover
+        <&> set envLogger lgr
+        <&> set envRegion defaultRegion
+  (runResourceT . runAWST env) awsAction
+
+overlappingRRS ∷ ResourceRecordSet → ResourceRecordSet → Bool
+overlappingRRS x y = x^.rrsName == y^.rrsName && x^.rrsType == y^.rrsType && x^.rrsResourceRecords == y^.rrsResourceRecords
```
domenkozar: (nitpick) let's split lines on `&&`

domenkozar:
Let's get rid of unicode syntax extension :)

domenkozar:
So NixOps figured out what zone to use on it's own when adding DNS entries?

deployments/cardano-nodes-config.nix
```diff
@@ -3,9 +3,4 @@ with (import ./../lib.nix);
 mergeNodes [
   (genNodes (range 0 1) (i: { region = "eu-central-1"; inherit i; }))
   (genNodes (range 2 3) (i: { region = "eu-west-1"; inherit i; }))
-  (genNodes (range 4 5) (i: { region = "eu-west-2"; inherit i; }))
```
domenkozar: This shouldn't be removed

deepfire: Yes, it was part of testing setup, meant to be removed before merging.

PR #65, opened 2017-06-13
[DEVOPS-153] fix cardano-node eval when nixops has not claimed an Elastic IP

cleverca22:


domenkozar:
As we discussed, let's rather expose the option in nixos modules and then populate it in nixops specific modules.

domenkozar:
Could you make a PR against staging? It changed a bit

PR #68, opened 2017-06-16
[DEVOPS-140] Remove developer access to machines

cleverca22:
* create a list of devops keys
* give devops access to the live-production user on cardano-deployer
* remove the main keylist from common.nix (which gave access to every machine)

it is still possible to access machines by running "nixops ssh" on the deployer


jmitchell:
I suspect the CI failure is because the authorized keys list was removed from `common.nix`. We may need at least one key there.

```
Failed assertions:
- Neither the root account nor any wheel user has a password or SSH authorized key.
You must set one to prevent being locked out of your system.
```

cleverca22:
which key do you think i should add back?

jmitchell:
Unless @domenkozar objects, I suggest using `devOpsKeys`. I don't anticipate scenarios where only a subset of the DevOps team needs access to a system.

cleverca22:
oh, and i know why the CI fails, the CI isn't aware of the extra root key nixops adds include the

domenkozar:
Yeah, let's circle back and have georgee as part of `devOpsKeys`

cleverca22:
done

PR #69, opened 2017-06-16
[DEVOPS-111]: change the restart timer to daily

cleverca22:
also change it from a cronjob to a systemd timer
and you can select between hourly and daily, with an adjustable rolling-restart gap

### DIFF
--- a/modules/cardano-node.nix
+++ b/modules/cardano-node.nix
@@ -23,7 +23,7 @@ let
 
   command = toString [
     cfg.executable
-    "--address ${if cfg.publicIP == null then "0.0.0.0" else cfg.publicIP}:${toString cfg.port}"
+    (optionalString (cfg.publicIP != null) "--address ${cfg.publicIP}:${toString cfg.port}")
     "--listen ${cfg.privateIP}:${toString cfg.port}"
     "--kademlia-address ${cfg.privateIP}:${toString cfg.port}"
     # Profiling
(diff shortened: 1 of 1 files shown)


modules/cardano-node.nix
```diff
@@ -167,6 +158,28 @@ in {
       allowedUDPPortRanges = [ { from = 1024; to = 65000; } ];
     };
 
+    # Workaround for CSL-1029
+    # Reboot cardano-node every day, offset by node id (in 4 minute intervals)
+    systemd.services.cardano-restart = let
+      getDailyTime = testIndex: let
+          # how many minutes between each node restarting
+          interval = 60;
+          minute = mod (testIndex * interval) 60;
```
jmitchell: Doesn't `minute` always end up as `0` here? If so, we might as well use a constant for clarity.

modules/cardano-node.nix
```diff
@@ -167,6 +158,28 @@ in {
...
+          hour = mod ((testIndex * interval) / 60) 24;
```
jmitchell: Similarly, `(testIndex * interval) / 60` given the constant interval of 60 will always be `testIndex`.

modules/cardano-node.nix
```diff
@@ -167,6 +158,28 @@ in {
...
+        in "${toString hour}:${toString minute}";
+      getHourlyTime = testIndex: let
+          # how many minutes between each node restarting
+          interval = 4;
+          minute = mod (testIndex * interval) 60;
+        in "*:${toString minute}";
+      getInterval = testIndex: if testIndex < 3 then (getDailyTime testIndex) else (getHourlyTime testIndex);
```
jmitchell: Where is `getInterval` used? Did you mean to use it instead of `getDailyTime` for the `startAt` attr below?

modules/cardano-node.nix
```diff
@@ -167,6 +158,28 @@ in {
...
+    in {
+      script = ''
+        /run/current-system/sw/bin/systemctl restart cardano-node
+      '';
+      startAt = getDailyTime cfg.testIndex;
+    };
+
```
jmitchell: How does this expression translate into a cron job? Perhaps the service definition for `cardano-restart` is missing?

modules/cardano-node.nix
```diff
@@ -167,6 +158,28 @@ in {
       allowedUDPPortRanges = [ { from = 1024; to = 65000; } ];
     };
 
+    # Workaround for CSL-1029
+    # Reboot cardano-node every day, offset by node id (in 4 minute intervals)
```
jmitchell: "4 minute intervals" may need to be changed.

modules/cardano-node.nix
cleverca22: i had originally written the code with interval=4; so its a question of keeping that flexibility or not

modules/cardano-node.nix
cleverca22: i was going to have nodes 0 and 1 restart daily, and keep the rest on hourly, but then domen said we can just switch all of staging to daily, thought i could leave it there to be flexible

modules/cardano-node.nix
cleverca22: [1] will auto-generate a systemd timer for every service with a .startAt attribute, which will start the service at the specified times
and with a time like "2:15", it would start the service every night at 2:15 am

[1] https://github.com/NixOS/nixpkgs/blob/release-17.03/nixos/modules/system/boot/systemd.nix#L803-L809

jmitchell: Ah, I see. Thanks for clarifying.

It's compatible with https://github.com/input-output-hk/iohk-nixops/blob/staging/modules/cardano-node.nix#L170-L194, right?

cleverca22: yeah, the service has a different name, so it wont directly conflict, it just relies on the name matching up between the systemctl invocation and the cardano-node definition

also, the old cronjob, and the new timer, may start cardano back up, both in the event of it failing to restart, and when it has been stopped

cleverca22:
https://gist.github.com/cleverca22/9c9619a42419ed190e3bfe417e6751aa

an example of the timestamps it can generate for various intervals

PR #70, opened 2017-06-19
[DEVOPS-123] nixpkgs unification

cleverca22:
most parts of the project now obey nixpkgs-src.json

the #!nix-shell stuff cant be configured to read from a file, so they now expect the user to `source scripts/set_nixpath.sh` before being ran, that will probably change after DEVOPS-22

still needs a travis check to ensure no new paths get added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ install:
   - mkdir keys
   - touch keys/key{0,1,2,3,4,5,6,7,8,9,10,11,12,13,41}.sk
 script:
-  - echo $NIX_PATH
+  - ./scripts/find-all-revisions.sh
   - nixops --version
   # check all scripts compile
   - ./CardanoCSL.hs --help
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,10 @@ language: nix
 
 install:
   - source scripts/set_nixpath.sh
-  - echo "binary-caches = https://cache.nixos.org https://hydra.iohk.io" | sudo tee -a /etc/nix/nix.conf
-  - echo "binary-cache-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=" | sudo tee -a /etc/nix/nix.conf
+  - cat /etc/nix/nix.conf > ~/nix.conf
+  - echo "binary-caches = https://cache.nixos.org https://hydra.iohk.io" >> ~/nix.conf
+  - echo "binary-cache-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=" >> ~/nix.conf
+  - export NIX_CONF_DIR=~
   - nix-env -iA nixopsUnstable -f '<nixpkgs>'
   - touch static/datadog-{api,application}.secret
   - echo "secret" > static/tarsnap-cardano-deployer.secret
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: nix
 
 install:
   - source scripts/set_nixpath.sh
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -1,15 +1,15 @@
 language: nix
 
-env:
-  NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/c0b1e8a5fb174cd405dcca9f7fec275714ad9f4b.tar.gz
-
 install:
+  - source scripts/set_nixpath.sh
   - nix-env -iA nixopsUnstable -f '<nixpkgs>'
```
deepfire: One particular idea (from https://github.com/input-output-hk/iohk-nixops/blob/devops-123-refactor/shell.nix#L45) is to:

1. use `shellHook` in a `shell.nix` to set `NIX_PATH`.
2. standartise on _not_ running anything outside of a nix shell

While it seems uncontroversial to do 1 _fully_, the extent to which we go with 2 is arguably open to debate.

I don't want to sound judgemental here -- but merely suggest this for discussion.

..and I think, that in today's standup meeting, when I brought this up (although the _nothing outside nix-shell_ wasn't worded strongly) Domen was in agreement.

What do you think?

cc @domenkozar

deepfire:
I'm running a slightly post-filtered version of the `grep` command from the issue:
`grep 'nixpkgs=\|nixpkgs/archive|nixpkgs\.git\|nixpkgs-channels\|nixpkgs-src' . -rI | grep -v '\./\.git/'`
and it gives me this extra thing:

- `./jobsets/spec.json`: `"nixpkgs": { "type": "git", "value": "https://github.com/NixOS/nixpkgs-channels.git nixos-16.09", "emailresponsible": false }`

cleverca22:
i have changed that one to match nixpkgs-src.json, and verified it passes tests on my own hydra

i need to look into the declarative jobs more, and see if i can make that automated

.travis.yml
cleverca22: also keep in mind, that the shellHook runs after nix-shell has eval'd a nix expression, so the nix file itself has to obey `nixpkgs-src.json` and not `NIX_PATH`

deepfire: @cleverca22, the way I've done this was:

- https://github.com/input-output-hk/iohk-nixops/blob/devops-123-refactor/shell.nix#L5, and
- https://github.com/input-output-hk/iohk-nixops/blob/devops-123-refactor/shell.nix#L45

cleverca22: yeah, we are going to have to find the good parts of both PR's and merge them together, i don't have any improvements to shell.nix

.travis.yml
```diff
@@ -1,15 +1,15 @@
...
   - touch static/datadog-{api,application}.secret
   - echo "secret" > static/tarsnap-cardano-deployer.secret
   - mkdir keys
   - touch keys/key{0,1,2,3,4,5,6,7,8,9,10,11,12,13,41}.sk
 script:
+  - source scripts/set_nixpath.sh
```
domenkozar: This shouldn't be needed, `install` directive already sets it.

cleverca22: my understanding was that travis might snapshot the vm after install is done, and then resume future builds at script

or is it just not that smart?

domenkozar: It's not :) Even if it did, it would preserve the env vars

cleverca22:
`The job exceeded the maximum time limit for jobs, and has been terminated.`
until we get the main iohk hydra building things, this job wont pass on hydra

domenkozar:
Something is fishy, all those haskell packages that are building should be used from the cache.

Source of trouble is `nix-instantiate jobsets/cardano.nix --show-trace`, we have to check what's going on.

cleverca22:
I've narrowed down why travis is mass-rebuilding, the new nixpkgs revision is causing all of pkgs/default.nix to rebuild, and those aren't being found in a binary cache, hydra.iohk.io clearly wont have it because the PR isnt getting merged or pre-built by PR #76 

are these supposed to align with the haskellPackages instances on hydra.nixos.org?

domenkozar:
I see, `shell.nix` is using `default.nix` packages.

@cleverca22 can you configure `nix.conf` on travis to use our binary cache?

cleverca22:
ah, that part got lost when i redid the PR, let me re-add it

cleverca22:
travis builds are timing out on here, but once #76 is merged, they should gain the support of the binary cache and pass

domenkozar:
1) needs to be rebased and conflicts resolved

2) missing grep check from AC

3) same changes against cardano-sl repository

domenkozar:
Needs the same PR for cardano-sl :)

PR #72, opened 2017-06-20
[DEVOPS-169] Initial NixOS tests for cardano-sl

cleverca22:


jobsets/tests/simple-node.nix
```diff
@@ -0,0 +1,22 @@
+import ./make-test.nix ({ pkgs, ... }: {
+  name = "boot";
+  nodes = {
+    machine = { config, pkgs, ... }: {
+      imports = [ ../../modules/cardano-node.nix ];
+      services.cardano-node = {
+        enable = true;
+        testIndex = 0;
```
domenkozar: Best to use `import modules/cardano-node-config.nix 0 null` under imports for now to avoid these settings be duplicated around

jobsets/tests/simple-node.nix
```diff
@@ -0,0 +1,22 @@
+import ./make-test.nix ({ pkgs, ... }: {
+  name = "boot";
```
domenkozar: Name should be `simple-node`

jobsets/tests/default.nix
```diff
@@ -0,0 +1,17 @@
...
+
+let
+  forAllSystems = genAttrs supportedSystems;
+  importTest = fn: args: system: import fn ({
+    inherit system;
+  } // args);
+  callTest = fn: args: forAllSystems (system: hydraJob (importTest fn args system));
+in rec {
+  # TODO: tests of images
+  boot = callTest ./boot.nix {};
```
domenkozar: Is there a good reason to have `boot.nix` around?

jobsets/tests/simple-node-nixops.nix
```diff
@@ -0,0 +1,22 @@
+import ./make-test.nix ({ pkgs, ... }: {
+  name = "boot";
+  nodes = {
+    machine = { config, pkgs, ... }: {
+      imports = [ ../../modules/cardano-node.nix <nixops/nix/options.nix> <nixops/nix/resource.nix> ];
```
domenkozar: Shouldn't this use then `cardano-node-aws.nix` if it's testing nixops? Otherwise nixops options aren't used.

But it will barf on `resources` not being populated, hmmm :)

jobsets/tests/make-test.nix
```diff
@@ -0,0 +1,5 @@
+f: { system ? builtins.currentSystem, ... } @args:
```
domenkozar: We could get rid of this file by calling `import <nixpkgs/nixos/tests/make-test.nix> ...` for each test

jobsets/tests/simple-node-nixops.nix
cleverca22: that was to test the code surrounding networking.privateIPv4, because it both exists, and doesn't exist, depending on if i import nixops/nix/options.nix
because nothing tries to use the values of those options, it doesn't fail out

jobsets/tests/default.nix
cleverca22: ah, that was just a test to make sure hydra was working before i added more

tests/simple-node-nixops.nix
```diff
@@ -0,0 +1,17 @@
+import <nixpkgs/nixos/tests/make-test.nix> ({ pkgs, ... }: {
+  name = "boot";
+  nodes = {
+    machine = { config, pkgs, ... }: {
+      imports = [ ../../modules/cardano-node-config.nix <nixops/nix/options.nix> <nixops/nix/resource.nix> ];
```
domenkozar: you'll need to call `(import ./../modules/cardano-node-config.nix 0 "")` or refactor the module not to be a sloppy function :)

cleverca22: ah, i was wondering why you had some random values, and why it broke

PR #81, opened 2017-06-23
also lock in the nixpkgs used for jobset eval

cleverca22:
PR #78 only covered the default.nix that generates jobsets

### DIFF
--- a/jobsets/cardano.nix
+++ b/jobsets/cardano.nix
@@ -1,9 +1,11 @@
-{ pkgs ? (import <nixpkgs> {}), supportedSystems ? [ "x86_64-linux" ] }:
+let
+  fixedNixpkgs = (import <nixpkgs> {}).fetchFromGitHub (builtins.fromJSON (builtins.readFile ../nixpkgs-src.json));
+in { pkgs ? (import fixedNixpkgs {}), supportedSystems ? [ "x86_64-linux" ] }:
 
 with pkgs;
 
 let
-  iohkpkgs = import ./../default.nix {};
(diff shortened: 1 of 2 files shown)


.nixpkgs-commit
```diff
@@ -0,0 +1 @@
+9b948ea439ddbaa26740ce35543e7e35d2aa6d18
```
domenkozar: Afaik this is not used?

jobsets/cardano.nix
```diff
@@ -1,9 +1,11 @@
-{ pkgs ? (import <nixpkgs> {}), supportedSystems ? [ "x86_64-linux" ] }:
+let
+  fixedNixpkgs = (import <nixpkgs> {}).fetchFromGitHub (builtins.fromJSON (builtins.readFile ../nixpkgs-src.json));
+in { pkgs ? (import fixedNixpkgs {}), supportedSystems ? [ "x86_64-linux" ] }:
```
domenkozar: Shouldn't this be called `nixpkgs` so it compiles with Hydra input `nixpkgs`?

cleverca22: when `jobsets/default.nix` generates a PR jobset, it will pass the wrong nixpkgs in, the one master once, not the one the PR wants

this will import the wrong one with `<nixpkgs>`, then use that to import the correct one

domenkozar:
@cleverca22 needs a rebase

cleverca22:
this change was also in #70, so i can just close this one

PR #82, opened 2017-06-23
[DEVOPS-32] ensure pkgs/default.nix is up to date

cleverca22:
found an edgecase in testing this, updating `pkgs/default.nix` changes the build of stack2nix, and now travis cant use the cached copy, so it needs a PR built by hydra

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ install:
   - touch keys/key{0,1,2,3,4,5,6,7,8,9,10,11,12,13,41}.sk
 script:
   - ./scripts/find-all-revisions.sh
+  - nix-shell -p cabal2nix stack cabal-install ghc -Q -j 4 --run scripts/check-stack2nix.sh
   - nixops --version
   # check all scripts compile
   - iohk-ops --help
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,14 @@ install:
   - echo "binary-caches = https://cache.nixos.org https://hydra.iohk.io" >> ~/nix.conf
   - echo "binary-cache-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=" >> ~/nix.conf
   - export NIX_CONF_DIR=~
-  - nix-env -iA nixopsUnstable -f '<nixpkgs>'
+  - nix-env -f '<nixpkgs>' -i -A nixopsUnstable -A git
   - nix-env -iA iohk-ops       -f 'default.nix'
   - touch static/datadog-{api,application}.secret
   - echo "secret" > static/tarsnap-cardano-deployer.secret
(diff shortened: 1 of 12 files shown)


cleverca22:
and 2 more edge cases,
* if the string `sudo` appears in `.travis.yml`, it will silently upgrade the builder from ubuntu 12.04 to ubuntu 14.04, and break nixops in the process: https://github.com/NixOS/nixops/issues/693

i also had an idea on how to perform stack2nix testing under hydra, just flag it as a fixed-output derivation, because we want the output to be "no differences found", and any deviation from that is a failure
* `stack2nix` calls `cabal2nix`, `cabal2nix` calls `nix-prefetch-url`, and `nix-prefetch-url` fails hard in a build, it cant access `/nix/var/db` and https://github.com/NixOS/nix/issues/13 is currently not an option

.travis.yml
```diff
@@ -5,6 +5,10 @@ env:
 
 install:
   - nix-env -iA nixopsUnstable -f '<nixpkgs>'
+  - echo "binary-caches = https://cache.nixos.org https://hydra.iohk.io" > ~/nix.conf
+  - echo "binary-cache-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=" > ~/nix.conf
+  - export NIX_CONF_DIR=~
```
domenkozar: This is a bad idea since the default file does bring some sanity things.

Try to workaround the 14.04 bug by doing `- export PYTHONPATH=`

jobsets/cardano.nix
```diff
@@ -17,4 +16,5 @@ in rec {
     ];
   }).config.system.build.amazonImage;
   tests = import ./../tests { inherit pkgs supportedSystems; };
+  #stack2nix-output = callPackage ./check-stack2nix.nix { inherit (iohkpkgs) stack2nix; };
```
domenkozar: This won't work since `stack2nix` is very impure, let's remove it from here :)

scripts/check-stack2nix.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+
+# Get relative path to script directory
+set -xe
+scriptDir=$(dirname -- "$(readlink -f -- "${BASH_SOURCE[0]}")")
+
+time nix-build -A stack2nix -o $scriptDir/stack2nix -Q -j 4
+
+time $scriptDir/stack2nix/bin/stack2nix --revision 4bcf1e6b601d531f753ebdb6aec23e19d4b08e6c https://github.com/input-output-hk/cardano-sl.git > $scriptDir/default-2.nix
```
domenkozar: I was hoping we could just run `pkgs/generate.sh` and check `git diff` to have no output. Then we wouldn't need this script at all. Am I missing something with that approach?

domenkozar:
@cleverca22 let's try rebasing this on top of latest master to see what happens :)

cleverca22:
@domenkozar all checks are now passing and the PR looks good to me

PR #85, opened 2017-06-27
disable the nixos image, it takes over an hour to build on aws

cleverca22:


domenkozar:
https://hydra.iohk.io/jobset/serokell/iohk-nixops-85 has built, but there's no status check

PR #88, opened 2017-06-29
Drop unsafe function from getNodesMap

imalsogreg:
Small change that removes a `read` and a `catMaybes` from config parsing. Misspellings of `DeploymentInfo` node names are returned in `Either String`, rather than (A) being dropped if they didn't start with `"node"` or (B) throwing an error if the suffix doesn't parse to an `Int`.

Type-checks but otherwise untested.

### DIFF
--- a/CardanoCSL.hs
+++ b/CardanoCSL.hs
@@ -17,6 +17,7 @@ import Control.Lens
 import Filesystem.Path.CurrentOS (encodeString)
 import Data.Aeson
 import Data.Aeson.Lens
+import qualified Data.List as L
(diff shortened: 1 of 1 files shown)


domenkozar:
Interesting, are you using this code somewhere? Just curious.

imalsogreg:
@domenkozar Not yet. I've been looking around for examples of how to deploy distributed systems with nix, which lead me to reading through this repo.

domenkozar:
Finally we just removed the function. Thanks for the patch and sorry for slowness :)

PR #99, opened 2017-07-06
[DEVOPS-194]  Fix scripts/aws.hs impurities:  make awscli specified as input

deepfire:
Tested on the `daedalus-installer-test` bucket.

https://issues.serokell.io/issue/DEVOPS-195#comment=96-5415


domenkozar:
Why did you commit `pkgs/default.nix`? Seems like it doesn't belong here and breaks CI

deepfire:
@domenkozar, that was by mistake.. fixed! : -)

PR #102, opened 2017-07-07
[DEVOPS-190] nixpkgs commit in config.yaml

deepfire:
This makes two changes (so far):

1. add nixpkgs commit to `NixopsConfig` and carry it from/to `config.yaml`
2. make the `template` command fill in its value from `nixpkgs-src.json`

TODO:

1. [ ] figure out how to make `iohk-ops deploy`/etc. respect that commit id value
    - hypothesis:  overwrite `nixpkgs-src.json`? (_/me laughs madly_)

### DIFF
--- a/iohk/NixOps.hs
+++ b/iohk/NixOps.hs
@@ -77,6 +77,12 @@ data NixGitSource
   } deriving (Generic, Show)
 instance FromJSON NixGitSource
 
+readNixGitSource :: String -> IO NixGitSource
+readNixGitSource path = do
+  bs <- BL.readFile path
+  pure $ flip fromMaybe (AE.decode bs) $
+    errorT $ format ("File doesn't parse as NixGitSource: "%s) (T.pack path)
+
 -- | The set of first-class types present in Nix
 data NixValue
   = NixBool Bool
@@ -388,9 +394,7 @@ generateGenesis :: Options -> NixopsConfig -> IO ()
 generateGenesis o c = do
   let cardanoSLSrcSpecFile = "cardano-sl-src.json"
       cardanoSLDir         = "cardano-sl"
-  bs    <- BL.readFile cardanoSLSrcSpecFile
-  let NixGitSource{..} = flip fromMaybe (AE.decode bs) $
-                         errorT $ format ("Version spec for 'cardano-sl' doesn't parse as NixGitSource: "%s) (T.pack cardanoSLSrcSpecFile)
+  NixGitSource{..} <- readNixGitSource cardanoSLSrcSpecFile
   printf ("Generating genesis using cardano-sl commit "%s%"\n") $ fromCommit rev
   exists <- testpath cardanoSLDir
   unless exists $
--- a/iohk/NixOps.hs
+++ b/iohk/NixOps.hs
@@ -245,13 +245,16 @@ parserOptions = Options
                 <*> switch  "serial"  's' "Disable parallelisation"
                 <*> switch  "verbose" 'v' "Print all commands that are being run"
 
+nixpkgsCommitPath :: Commit -> Text
+nixpkgsCommitPath = ("nixpkgs=" <>) . fromURL . nixpkgsNixosURL
+
 nixopsCmdOptions :: Options -> NixopsConfig -> [Text]
 nixopsCmdOptions Options{..} NixopsConfig{..} =
   ["--debug"   | oDebug]   <>
   ["--confirm" | oConfirm] <>
   ["--show-trace"
   ,"--deployment", cName
-  ,"-I", "nixpkgs=" <> (fromURL $ nixpkgsNixosURL cNixpkgsCommit)
+  ,"-I", nixpkgsCommitPath cNixpkgsCommit
   ]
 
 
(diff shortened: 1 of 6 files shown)


domenkozar:
No, `iohk-ops` should read from the configuration file it has been passed under `-c`

deepfire:
@domenkozar, it already does _read_ it -- the question is how best effectuate the value that is read.

domenkozar:
It should pass `nixops deploy -I nixpkgs=...` when calling nixops

deepfire:
@domenkozar, that raises following questions:

1. if we just reference the `nixpkgs` tarball blob, then we lose the benefits of caching it..
2. we have `nixpkgs-src.json` parametrising nodes directly via `modules/common.nix` -- if we just pass `-I` to `nixops deploy` (and the myriad others), then.. it would seem to me we'll have an inconsistency

iohk/CardanoCSL.hs
```diff
@@ -125,16 +125,6 @@ dumpLogs o c withProf nodes = do
              , defLogs
              ]
 
-generateIPDHTMappings :: Options -> NixopsConfig -> IO Text
```
domenkozar: This shouldn't be dropped since it's useful for development and testing a feature with a custom installer.

domenkozar:
Now that https://github.com/input-output-hk/iohk-nixops/pull/107 is merged, we need to just do:

    $ NIX_PATH_LOCKED nixops deploy -I https://....

cleverca22:
i think a wrapper around `nixops deploy --build-only` would also be of use, so i could test building the entire deployment without having to give it an accesskey and accidentally creating machines

domenkozar:
@cleverca22 sounds good, but let's keep that out of scope of this issue

deepfire:
@domenkozar, if we set NIX_PATH to point to a tarball, we lose the benefits of caching nixpkgs themselves.

domenkozar:
That's premature optimization, we're talking about a few seconds of time for a deploy that takes 1h+

deepfire:
@domenkozar, two points:

1. we seem have two commands that depend on nixpkgs -- `nixops modify` and `nixops deploy` -- maybe I'm wrong and only one of them depends, I'm not sure yet
2. you are entirely right, if we are talking about successful production deployments. However this kills the interactive development process -- which is naturally iterative, since Nix is untyped

deepfire:
Just looked up -- both `modify` and `deploy` take `-I`.

While this isn't a proof that both must have proper `nixpkgs` set, this is a strong hint..

So we are talking _two_ tarball fetches per `iohk-ops deploy`.

cleverca22:
if the same URL is given to both, nix can cache the fetched tarball and reuse it

deepfire:
Proceeding, then.

deepfire:
Ha, the fun thing is how Nix reacts to an inexistent version of nixpkgs:

```
nix-shell:~/devops-190]$ iohk-ops -c config.yaml do deploy
Deploying cluster devops-190
downloading ‘https://github.com/NixOS/nixpkgs/archive/1b948ea439ddbaa26740ce35543e7e35d2aa6d18.tar.gz’... [0/0 KiB, 0.0 KiB/s]
warning: Nix search path entry ‘https://github.com/NixOS/nixpkgs/archive/1b948ea439ddbaa26740ce35543e7e35d2aa6d18.tar.gz’ cannot be downloaded, ignoring
node1......................> waiting for IP address...
node0......................> waiting for IP address...
node1......................> [running]
```

I.e. it basically ignores the wrongdoing.

deepfire:
@domenkozar, ready for review.

lib.nix
```diff
@@ -29,9 +29,10 @@ in lib // (rec {
     ) nodes;
 
   # fetch nixpkgs and give the expected hash
+  fetchNixpkgsWithNixpkgs = nixpkgs: nixpkgs.fetchgit (removeAttrs (builtins.fromJSON (builtins.readFile ./nixpkgs-src.json)) ["date"]);
```
domenkozar: fetchgit is very slow for nixpkgs

deepfire:
@domenkozar, things done since last review:

1. default `-c` to `config.yaml`, instead of failing
2. resurrect `io generate-ipdht`
3. drop the commented out code

deepfire:
4. restore the `fetchFromGitHub` semantics of `nixpkgs-src.json`
5. make the thing fail on wrong Nixpkgs

iohk/NixOps.hs
```diff
@@ -76,6 +98,21 @@ data NixGitSource
...
 instance FromJSON NixGitSource
+instance ToJSON NixGitSource
+
+readNixGitSource :: Project -> IO NixGitSource
+readNixGitSource (projectSrcFile -> path) = do
+  bs <- BL.readFile (T.unpack $ format fp path)
+  pure $ flip fromMaybe (AE.decode bs) $
+    errorT $ format ("File doesn't parse as NixGitSource: "%fp) path
+
+writeNixGitSource :: Project -> NixGitSource -> IO ()
```
domenkozar: Unused function

domenkozar:
I'm not too happy about changing cli interface as we need to change docs and habits, but I guess it's still early for `iohk-ops`.

deepfire:
Yes, I agree -- I shall try to put more thought into picking command structure in the future.

domenkozar:
Some legit CI failures:

```
$ iohk-ops template --here                          --config 'csl.yaml'                  'csl'                  Nodes Explorer ReportServer
iohk-ops: File doesn't parse as NixGitSource: nixpkgs-src.json
CallStack (from HasCallStack):
  error, called at ./NixOps.hs:574:10 in main:NixOps
```

PR #103, opened 2017-07-08
save coredumps from cardano-sl

cleverca22:


modules/cardano-node.nix
```diff
@@ -187,6 +187,7 @@ in {
         key = keyId + ".sk";
       in ''
         [ -f /run/keys/${keyId} ] && cp /run/keys/${keyId} ${stateDir}${key}
+        ulimit -c unlimited
```
domenkozar: Why is this needed? Would be good to have a comment. Is it save to have it on for production?

modules/common.nix
```diff
@@ -24,6 +24,8 @@ in {
 
   environment.variables.TERM = "xterm-256color";
 
+  systemd.coredump.enable = true;
```
domenkozar: Is it safe to have coredump happen in production?

modules/cardano-node.nix
cleverca22: systemd.coredump.enable only sets a hook that captures cores, ulimit is still required to enable any given program to dump

modules/common.nix
cleverca22: hmmm, it will contain copies of the private key data in the cores

domenkozar:
@cleverca22 sadly `productionMode` is a different thing than the production environment. I think we want to enable these only for staging environment (which are the files suffixed with -env-staging.nix`

PR #109, opened 2017-07-15
start of darwin build jobs

cleverca22:


cleverca22:
found a surprise in the darwin jobset stuff
```
socket-io = super.callCabal2nix "socket-io" "${socket-io-src}/socket-io" {};
```
this uses import from derivation to run `cabal2nix` and import it, and when set to do darwin builds, it uses the darwin `cabal2nix`, and hydra doesn't allow build slaves at eval time!

modules/hydra-master.nix
```diff
@@ -10,6 +10,13 @@ let
     system = "x86_64-linux";
     supportedFeatures = [ "kvm" "nixos-test" ];
   };
+  mkMac = hostName: commonBuildMachineOpt // {
+    inherit hostName;
+    maxJobs = 1;
```
domenkozar:
```
DE302:~ admin$ sysctl -n hw.ncpu
6
```

I think we should go at least with 2.

PR #110, opened 2017-07-17
rename the hydra jobsets

cleverca22:
add a jobset to build cardano-sl master
add a dummy PR json to aid in testing

### DIFF
--- a/jobsets/default.nix
+++ b/jobsets/default.nix
@@ -1,4 +1,4 @@
-{ nixopsPrsJSON, cardanoPrsJSON, nixpkgs ? <nixpkgs>, declInput ? {} }:
+{ nixopsPrsJSON ? ./simple-pr-dummy.json, cardanoPrsJSON ? ./simple-pr-dummy.json, nixpkgs ? <nixpkgs>, declInput ? {} }:
 
 # Followed by https://github.com/NixOS/hydra/pull/418/files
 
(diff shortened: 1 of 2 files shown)


jobsets/default.nix
```diff
@@ -1,4 +1,4 @@
-{ nixopsPrsJSON, cardanoPrsJSON, nixpkgs ? <nixpkgs>, declInput ? {} }:
+{ nixopsPrsJSON ? ./simple-pr-dummy.json, cardanoPrsJSON ? ./simple-pr-dummy.json, nixpkgs ? <nixpkgs>, declInput ? {} }:
```
domenkozar: Why are these defaults needed?

cleverca22: they make testing outside of hydra far simpler, you can just nix-build the file and get json you can diff

domenkozar:
I think we also need to change hydra.conf configuration for status updates

cleverca22:
https://github.com/input-output-hk/iohk-nixops/blob/master/modules/hydra-master.nix#L54

its configured to notify for any jobset that has an input called `jobsets`

domenkozar:
@cleverca22 but doesn't `jobs = serokell:*` filter everything by the serokell project?

cleverca22:
yeah, that will have to be fixed when we rename the project

PR #111, opened 2017-07-24
Miscellanea 0

deepfire:
As per https://issues.serokell.io/issue/DEVOPS-139#comment=96-5809

cc @domenkozar


deepfire:
A full deployment tested successfully.

default.nix
```diff
@@ -75,5 +69,14 @@ let
                     --prefix PATH : "${pkgs.lib.makeBinPath iohk-ops-extra-runtime-deps}"
                   '';
                });
+
+    #mkDerivation = args: super.mkDerivation (args // {
+    #enableLibraryProfiling = false;
+    #});
+  };
+});
+  cabal2nixpkgs = rec {
```
domenkozar: Why has `iohk-ops` been left out? It uses `compiler.callPackage`, so nothing from `pkgs/default.nix`

iohk/NixOps.hs
```diff
@@ -525,17 +534,30 @@ scpFromNode o c (fromNodeName -> node) from to = do
     ExitFailure code -> TIO.putStrLn $ "scp from " <> node <> " failed with " <> showT code
 
 sshForEach :: Options -> NixopsConfig -> [Text] -> IO ()
-sshForEach o c cmd = 
+sshForEach o c cmd =
   nixops o c "ssh-for-each" ("--": cmd)
 
+deployed'commit :: Options -> NixopsConfig -> NodeName -> IO ()
```
domenkozar: Can you show the purpose of this command? I'd prefer not to have such hacks, but I'd like to understand the motivation as there might be an easier way.

iohk/iohk-ops.hs
```diff
@@ -11,20 +11,28 @@ import qualified Filesystem.Path.CurrentOS     as Path
...
 
 

 -- * Elementary parsers
 --
-optReadLower :: Read a => ArgName -> ShortName -> Optional HelpMessage -> Parser a
-optReadLower = opt (readMaybe . T.unpack . T.toTitle)
-argReadLower :: Read a => ArgName -> Optional HelpMessage -> Parser a
-argReadLower = arg (readMaybe . T.unpack . T.toTitle)
+diagReadMaybe :: (Bounded a, Enum a, Read a, Show a) => String -> Maybe a
```
domenkozar: This seems fairly complicated, why don't we use our own `Read` instance instead?

iohk/NixOps.hs
deepfire: @domenkozar, basically this came up when Philipp was working with the cluster and he doubted that the actually running version was what he wrote in `cardano-sl-src.json`.

This is not a hack, I believe -- this is a _sensor_, a debug tool to figure out the actual reality of what was deployed on a particular cluster.

modules/cardano-explorer.nix
```diff
@@ -1,3 +1,10 @@
+let
+
+  explorer' = (import ./../default.nix {}).cardano-sl-explorer-static;
+  explorer  = builtins.trace "using cardano-sl-explorer ${explorer'}" explorer';
```
domenkozar: `trace` shouldn't be used to pollute the deployment log. The source of information what is going to be deployed is at various places, like closure of the machine or `cardano-sl-explorer-src.json`. But we should come up with why we need this information printed and then what tooling to create around it.

default.nix
deepfire: @domenkozar, it uses `iohk-ops-extra-runtime-deps` which uses `self.yaml`, to avoid pulling in a set of different haskell packages.

modules/cardano-node.nix
```diff
@@ -6,7 +6,8 @@ let
   cfg = config.services.cardano-node;
   name = "cardano-node";
   stateDir = "/var/lib/cardano-node/";
-  cardano = (import ./../default.nix { inherit pkgs; }).cardano-sl-static;
+  cardano' = (import ./../default.nix { inherit pkgs; }).cardano-sl-static;
+  cardano  = builtins.trace "using cardano-sl ${cardano'}" cardano';
```
domenkozar: Same as for explorer.

shell.nix
```diff
@@ -56,6 +56,14 @@ mkDerivation {
     export NIX_PATH=nixpkgs=${nixpkgs}
     export NIX_PATH_LOCKED=1
     echo   NIX_PATH set to $NIX_PATH >&2
+
+    alias io-csl-path-commit='echo -n "Enter cardano-sl nix store path: "; read csl_path; nix-store --query --references $(nix-store --query --deriver $csl_path) | grep cardano-sl- | grep -v sl-core\\\|sl-db\\\|sl-infra\\\|sl-lrc\\\|sl-ssc\\\|sl-godtossing\\\|sl-txp\\\|sl-update | cut -c56-62'
```
domenkozar: This duplicates the new command `dump-commits`, we ought to have one code to show the versions/commits used.

modules/cardano-explorer.nix
deepfire: The answer is that this actually came up when a developer doubted that things were deployed as we claim them to be.

shell.nix
```diff
@@ -56,6 +56,14 @@ mkDerivation {
...
+    alias git-rev='git log -n1 --pretty=oneline'
```
domenkozar: This alias is a preference of style, for example, I use `tig`. I prefer we add those aliases in `.bashrc` and people can use them on the deployment machines.

config.nix
```diff
@@ -8,6 +8,7 @@ rec {
   # Run time settings
   bitcoinOverFlat = false;
   totalMoneyAmount = 60000000;
+  richPoorDistr = false;
```
domenkozar: Why has this flag been added?

iohk/iohk-ops.hs
deepfire: Well, that would add a fair number of boilerplate lines.

shell.nix
deepfire: We can change it to `tig` -- the problem with `~/.bashrc` is that it is different everywhere, whereas `nix-shell` travels with us and establishes a standard environment.

config.nix
deepfire: Because a development evaluation breaks if isn't defined at all.  `false` is the default value, as far as I can tell, so this should be a no-op.

modules/cardano-explorer.nix
deepfire: See the explanation above.

modules/cardano-node.nix
deepfire: See the explanation above.

modules/cardano-explorer.nix
deepfire: But maybe this isn't needed, if we choose the `deployed-commit` command.

modules/cardano-node.nix
deepfire: But maybe this isn't needed, if we choose the `deployed-commit` command.

shell.nix
deepfire: Yes, this isn't needed, if we choose the `deployed-commit` command.

default.nix
```diff
@@ -55,17 +57,9 @@ let
...
-  };
-});
-  iohk-ops-extra-runtime-deps = [
-    pkgs.git pkgs.nix-prefetch-scripts
-  ];
-  cabal2nixpkgs = rec {
-    # extra packages to expose, that have no relation to pkgs/default.nix
-    stack2nix = compiler.callPackage ./pkgs/stack2nix.nix {};
+    iohk-ops-extra-runtime-deps = [
+      pkgs.git pkgs.nix-prefetch-scripts self.yaml
```
domenkozar: self.yaml -> haskellPackages.yaml

pkgs/iohk-ops.nix
```diff
@@ -5,7 +5,7 @@
 mkDerivation {
   pname = "iohk-ops";
   version = "0.1.0.0";
-  src = ../iohk;
+  src = ./iohk;
```
cleverca22: reverting this change should fix the hydra eval

shell.nix
```diff
@@ -1,65 +1,34 @@
...
+let
+  localLib = import ./lib.nix;
+  nixpkgs  = localLib.fetchNixPkgs;
+  pkgs     = import nixpkgs {};
+  iohkpkgs = import ./default.nix {};
+in
+
+{ ... }:
+let
+  default = import ./pkgs/iohk-ops.nix;
```
cleverca22: no need to import, callPackage can take a path and import for you

pkgs/iohk-ops.nix
deepfire: Thank you @cleverca22!

iohk/NixOps.hs
```diff
@@ -348,28 +349,34 @@ readConfig cf = do
...
 parallelIO Options{..} =
   if oSerial
   then sequence_
   else sh . parallel
 
+logCmd  cmd args = do
+  printf ("-- "%s%"\n") $ T.intercalate " " $ cmd:args
+  Sys.hFlush Sys.stdout
+
+inproc :: Text -> [Text] -> Shell Line -> Shell Line
```
domenkozar: This function is not used anywhere?

modules/cardano-explorer.nix
```diff
@@ -1,3 +1,9 @@
+let
+
+  explorer = (import ./../default.nix {}).cardano-sl-explorer-static;
```
domenkozar: This is not used?

pkgs/iohk-ops.nix
domenkozar: Is it reverted?

shell.nix
domenkozar: cc @deepfire

pkgs/iohk-ops.nix
deepfire: Yes, now it is..

iohk/NixOps.hs
deepfire: It might not be in this branch, but might be in other -- it's sort of a symmetric element in a diamond of various options of calling shell stuff.

modules/cardano-explorer.nix
deepfire: It is now.

domenkozar:
Needs `pkgs/generate.sh` run due to upstream changes, otherwise looks good.

deepfire:
Ok, I've partially re-tested some of the most-well-trodden paths, and it seems to work.

Merging, as soon as it passes CI.

deepfire:
I can't seem to be able to restart a build -- Travis reports a successful restart, once I ask it, but that's all it ends up doing, it seems.

deepfire:
Hah, a force push did help where _restart build_ couldn't..

PR #118, opened 2017-08-01
Cryptonite: remove pthread linker flag

domenkozar:
I'm trying to test if we really still need this flag and how to fix it properly upstream.


domenkozar:
Submitted upstream: https://github.com/haskell-crypto/cryptonite/pull/184

We can get rid of the override once cryponite 0.24.1 is released.

PR #120, opened 2017-08-04
DON'T MERGE: try using older rev of stack2nix to fix Hydra builds

jmitchell:
./pkgs/generate.sh ends up producing the same nix expression, so this
is solely for documentation purposes while we diagnose the fault.


domenkozar:
Afaik we can close this now?

PR #125, opened 2017-08-16
iohk-ops | set-rev nixpkgs:  specialise by using nix-prefetch-url --unpack

deepfire:


iohk/NixOps.hs
```diff
@@ -1,6 +1,6 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
-{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FlexibleContexts, FlexibleInstances #-}
```
domenkozar: Each pragma is in it's own line

deepfire:
@domenkozar:
```
[nix-shell:~/iohk]$ io set-rev nixpkgs f40f2f21c03bec1cabd2a59c1863659e8c4d60dc
Changing nixpkgs commit from 9b948ea439ddbaa26740ce35543e7e35d2aa6d18 to f40f2f21c03bec1cabd2a59c1863659e8c4d60dc
downloading ‘https://github.com/NixOS/nixpkgs/archive/f40f2f21c03bec1cabd2a59c1863659e8c4d60dc.tar.gz’... [10244/11410 KiB, 1682.1 KiB/s]
unpacking...
path is ‘/nix/store/9wqw5m4yll4hbf08948jkcij7ia2hi5q-f40f2f21c03bec1cabd2a59c1863659e8c4d60dc.tar.gz’

[nix-shell:~/iohk]$ git diff
diff --git a/nixpkgs-src.json b/nixpkgs-src.json
index c16d090..96f47e2 100644
--- a/nixpkgs-src.json
+++ b/nixpkgs-src.json
@@ -1,6 +1 @@
-{
-    "owner":  "NixOS",
-    "repo":   "nixpkgs",
-    "rev":    "9b948ea439ddbaa26740ce35543e7e35d2aa6d18",
-    "sha256": "17208in65j29xsipzfdy5hi0nmqf836i28bkcihh18qys7669bj8"
-}
+{"owner":"NixOS","repo":"nixpkgs","sha256":"03nmgr3dyc4ah2vx047snhkx5nxafmh3syvyzvr4qjwk4aqxmmjd","rev":"f40f2f21c03bec1cabd2a59c1863659e8c4d60dc"}
\ No newline at end of file
```

deepfire:
I've tried re-enter the `nix-shell`, with the changed nixpkgs source definitions, and found that somehow the prefetched store path isn't used -- so the source is re-fetched.  But on the next invocation it's cached good.

iohk/NixOps.hs
```diff
@@ -73,6 +75,9 @@ projectURL     IOHK            = "https://github.com/input-output-hk/iohk-nixops
 projectURL     Nixpkgs         = "https://github.com/nixos/nixpkgs.git"
 projectURL     Stack2nix       = "https://github.com/input-output-hk/stack2nix.git"
 
+nixpkgsRevTarball :: Commit -> URL
+nixpkgsRevTarball (Commit rev) = URL $ "https://github.com/NixOS/nixpkgs/archive/"<>rev<>".tar.gz"
```
domenkozar: Let's also add in `owner` and `repo` so in case we fork.

deepfire: I'm feeling wary of doing anything more involved at this current moment, because then we'd have resolve merge conflicts with `staging`.  The proper way would be to extend the URL data type to allow more structured options, and tie that into defaults and optional args.

But that's going to get problematic during merge conflict resolution.

domenkozar: Why can't we just do `"https://github.com/" <> owner <> "/" <> repo <> "/archive/"<>rev<>".tar.gz"`

deepfire: That would make us enter all of those values for all repositories served by `set-rev`, and to me that looks highly unergonomic.

To resolve that we'd need defaulting.  To do defaulting we'd need to extend the URL type.

deepfire: But seriously, that's minutiae, isn't it?

domenkozar: I don't understand what you mean, can you elaborate?

deepfire: 1. the `set-rev` command currently accepts a revision argument
2. it takes the GH owner and repo implied from the default URLs for the projects it's set-revving -- which are basically strings now
3. if we choose to parametrise owner and repo, we either require them (bad) or provide defaults (good)
4. the good option is more involved

domenkozar: If I'm reading code correctly, it takes owner/repo from `nixpkgs-src.json` and and then writes it back.

Currently, if someone would change the owner to `input-output-hk` and bump the revision, it would read the json file, then fetch from hardcoded URL and fail since commit wouldn't be in `NixOS` repo.

I'm asking if we can use the information read from the JSON as truth and then only change the revision as per CLI options.

domenkozar: @deepfire ping

iohk/iohk-ops.hs
```diff
@@ -282,7 +283,14 @@ runTemplate o@Options{..} Template{..} = do
 runTemplate Options{..} _ = error "impossible"
 
 runSetRev :: Options -> Project -> Commit -> IO ()
-runSetRev o proj rev = do
+runSetRev o Nixpkgs newRev = do
+  oldGHSrc@Ops.GithubSource{..} <- Ops.readSource Ops.githubSource Nixpkgs
+  printf ("Changing nixpkgs commit from "%s%" to "%s%"\n") (fromCommit ghRev) (fromCommit newRev)
+  newSha256 <- incmd o "nix-prefetch-url" ["--unpack", fromURL $ Ops.nixpkgsRevTarball newRev]
```
domenkozar: Due to https://github.com/input-output-hk/iohk-ops/pull/129 we should remove `--unpack` and test if the two PRs work together.

domenkozar:
This has been since merged to staging, while the review wasn't addressed.

deepfire:
Yeah, @domenkozar, sadly, I've missed the github notifications.

Getting swamped by them.. : -/

PR #129, opened 2017-08-20
[DEVOPS-333] allow the nix expressions to work with a blank NIX_PATH

cleverca22:
this change allows the nix expressions to work with a blank NIX_PATH variable, so there is no chance for impurities to leak in

once tested under such a situation, it is safe to ignore NIX_PATH, and just leave it as-is on any system

todo:
* [ ] update `iohk-ops set-rev` to handle the new hash for nixpkgs (the tar hash, rather then the unpacked hash)
* [ ] update stack2nix to do the same: https://github.com/input-output-hk/stack2nix/pull/59

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -2,8 +2,14 @@
 # nix-repl lib.nix
 
 let
-  hostPkgs = import <nixpkgs> {};
-  lib = hostPkgs.lib;
+  # Allow overriding pinned nixpkgs for debugging purposes via iohkops_pkgs
+  fetchNixPkgs = let try = builtins.tryEval <iohkops_pkgs>;
+    in if try.success
+    then builtins.trace "using host <iohkops_pkgs>" try.value
+    else import ./fetch-nixpkgs.nix;
+
+  pkgs = import fetchNixPkgs {};
+  lib = pkgs.lib;
 in lib // (rec {
   ## nodeElasticIP :: Node -> EIP
   nodeElasticIP = node:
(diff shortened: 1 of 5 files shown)


lib.nix
```diff
@@ -2,8 +2,39 @@
...
+    args = [ (builtins.toFile "builder" script) ];
+    inherit input;
+    coreutils = builtins.storePath nixcfg.coreutils;
+    tar = builtins.storePath nixcfg.tar;
+    gzip = builtins.storePath nixcfg.gzip;
+  };
+  fetchNixpkgsTarWithNix = let
+    spec = builtins.fromJSON (builtins.readFile ./nixpkgs-src.json);
+  in import <nix/fetchurl.nix> {
+    url = "https://github.com/${spec.owner}/${spec.repo}/archive/${spec.rev}.tar.gz";
```
domenkozar: Did you try `unpack = true;`?

cleverca22: that only supports NAR files

cleverca22: https://github.com/NixOS/nix/blob/master/src/libstore/builtins.cc#L56-L61
nar, optionally wrapped in xz, and no other formats allowed

lib.nix
```diff
@@ -2,8 +2,13 @@
 # nix-repl lib.nix
 
 let
-  hostPkgs = import <nixpkgs> {};
-  lib = hostPkgs.lib;
+  fetchNixPkgs = let try = builtins.tryEval <nixops_pkgs>;
+    in if try.success
+    then builtins.trace "using host <nixops_pkgs>" try.value
+    else builtins.trace "fetching nixpkgs" (import ./fetch-nixpkgs.nix);
```
domenkozar: I think in `else` case we can leave out the trace since it's a desired operation.

domenkozar:
@cleverca22 `undefined variable ‘fetchNixpkgsWithNixpkgs’ at /home/travis/build/input-output-hk/iohk-ops/modules/common.nix:56:18`

PR #130, opened 2017-08-23
Staging

deepfire:
This delivers the core/relay split with up-to-date topology, explorer and report-server.

cc @domenkozar 

Review checklist:

- [x] merge `master` branch into `staging`
- [x] merge `staging-federated` into `staging`
- [x] make CI pass
- [x] drop the deployerIP from the Nixops deployment config file
- [x] make development environment Elastic IP-free (depends on https://issues.serokell.io/issue/DEVOPS-320) & re-enable it in CI
  - [x] switch develoment environment to an allow-all SG
- [x] ~~switch `deploy-full` to local checkouts~~ (non-pushing to both `cardano-sl` and `iohk-ops` contradicts having Hydra builds, making this undesirable)
- [x] ~~repair Hydra's `simpleNodeTest` & its sibling~~ (the tests are to be moved to the Cardano repository)
- [ ] *WIP* ensure review checklist is complete (https://issues.serokell.io/issue/DEVOPS-420)


iohk/NixOps.hs
```diff
@@ -403,156 +626,374 @@ exists o c@NixopsConfig{..} = do
...
     ExitSuccess -> return ()
     ExitFailure code -> TIO.putStrLn $ "scp from " <> node <> " failed with " <> showT code
 
 sshForEach :: Options -> NixopsConfig -> [Text] -> IO ()
-sshForEach o c cmd =
-  nixops o c "ssh-for-each" ("--": cmd)
+sshForEach o c command =
+  nixops o c "ssh-for-each" $ Arg <$> "--": command
+
+deployed'commit :: Options -> NixopsConfig -> NodeName -> IO ()
```
domenkozar: Nope

iohk/NixOps.hs
```diff
@@ -403,156 +626,374 @@ exists o c@NixopsConfig{..} = do
...
+    echo "User declined to proceed, exiting."
+    exit $ ExitFailure 1
+
+wipeNodeDBs :: Options -> NixopsConfig -> IO ()
+wipeNodeDBs o c@NixopsConfig{..} = do
+  confirmOrTerminate "Wipe node DBs on the entire cluster?"
+  parallelSSH o c "rm" ["-rf", "/var/lib/cardano-node"]
+  echo "Done."
+
+updateNixops :: Options -> NixopsConfig -> IO ()
```
domenkozar: Instead of this, we should just use `nix-build -A nixops` and modify what's there to be used for deployments

lib.nix
```diff
@@ -5,35 +5,41 @@ let
...
-
-  # Given a list of integers and a function,
-  # generate an attribute set with that many nodes and call the function with node index
-  genNodes = nodeLimit: ids: f:
-             genAttrs' (lib.flip builtins.filter ids
-                          (id: id < nodeLimit))
-                       (i: "node${toString i}") f;
+  debugBuild  = let x = (builtins.getEnv "DEBUG_BUILD")  != "";
+                in if !x then x else builtins.trace "DEBUG_BUILD is set, switching on debug build flags" x;
+  debugDeploy = let x = (builtins.getEnv "DEBUG_DEPLOY") != "";
```
domenkozar: We can just always enable those packages to be available, no need for a flag.

lib.nix
```diff
@@ -5,35 +5,41 @@ let
...
-  genAttrs' = names: fkey: fname:
-    lib.listToAttrs (map (n: lib.nameValuePair (fkey n) (fname n)) names);
-
-  # Given a list of integers and a function,
-  # generate an attribute set with that many nodes and call the function with node index
-  genNodes = nodeLimit: ids: f:
-             genAttrs' (lib.flip builtins.filter ids
-                          (id: id < nodeLimit))
-                       (i: "node${toString i}") f;
+  debugBuild  = let x = (builtins.getEnv "DEBUG_BUILD")  != "";
```
domenkozar: What's the difference in size of the binaries when we use dontStrip vs doStrip?

deployments/cardano-nodes-config.nix
```diff
@@ -1,12 +1,160 @@
...
+            protocol = "tcp";
+            fromPort = 8080; toPort = 8080;
+            sourceIp = "0.0.0.0/0";
+          }];
+        };
+      };
+    coreSGNames = core:
+        [ "allow-cardano-static-peers-${core.name}-${core.value.region}" ];
+    coreSGs     = nodePort: ips: core:
+      let neighbourNames = traceSF (p: "${core.name} peers: " + concatStringsSep ", " core.value.peers) core.value.peers;
```
domenkozar: No use of traceSF by default, it creeps the output

config-system-start.nix
```diff
@@ -0,0 +1 @@
+1502833071
```
domenkozar: This is not used anywhere?

deployments/cardano-nodes-config.nix
```diff
@@ -1,12 +1,160 @@
...
+                          in if xs != [] then builtins.elemAt xs 0
+                             else throw "No indexedNode by name '${name}'.";
+    # Canonical node parameter format is:
+    #   i             :: Int
+    #   region        :: String
+    #   type          :: String               -- one of: 'core', 'relay'
+    #   static-routes :: [['nodeId, 'nodeId]] -- here we go, TupleList..
+    #
+    # WARNING: this depends on the sort order, as explained above.
+    firstRelay      = findFirst (x: x.value.type == "relay") (throw "No relays in topology, it's sad we can't live..") indexed;
```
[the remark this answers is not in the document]
deepfire: There goes the Bladerunner allusion.. : -)

deployments/cardano-nodes-config.nix
deepfire: @domenkozar, it's arguably critical information -- we're now having more steps in the pipeline than we had before -- and it's getting worse.

Having visibility helps..

But if you're feeling strongly about it, of course we can dump it.

deployments/cardano-nodes-env-development.nix
```diff
@@ -0,0 +1,11 @@
+{ accessKeyId, deployerIP, systemStart, environment, ... }:
+
+with (import ./../lib.nix);
+let
+  nodeArgs = (import ./cardano-nodes-config.nix { inherit accessKeyId deployerIP systemStart  environment; }).nodeArgs;
+  nodeConf = import ./../modules/cardano-node-development.nix;
+in {
+  resources = rec {
+    elasticIPs = mkNodeIPs nodeArgs accessKeyId;
```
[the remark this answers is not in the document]
deepfire: Right, that's a left-over..

iohk/Topology.hs
```diff
@@ -0,0 +1,246 @@
+--  Imported from cardano-sl://infra/Pos/Network/Yaml.hs
```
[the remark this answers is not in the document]
deepfire: 1. `iohk-ops` needs to parse the topology:
   - to support iteration over specific node classes
   - to provide a stable and simplified output for Nix
   - to parametrise genesis `M` parameter
2. having this code in Cardano would make it harder to guarantee consistent topology->Nix interface -- usual cross-repository questions arise
3. having the parsing code maintained in `iohk-ops` means we can choose our own pace at which to adopt new changes

It wasn't an easy decision to make for me.

iohk/iohk-env.nix
```diff
@@ -0,0 +1,21 @@
+# Use as follows:
```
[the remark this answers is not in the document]
deepfire: The purpose of this file is to allow installation a version of GHC with packages required to run `iohk-ops`.

This makes `iohk-ops` independent of any particular `nix-shell`, which is very helpful.

lib.nix
deepfire: I didn't look into that..

lib.nix
deepfire: `tcpdump` is a huge parser written in C, a memory-unsafe language -- it tends to have RCE vulnerabilities. : -/

I'd rather us not have vulnerable binaries on production machines..

lib.nix
```diff
@@ -5,35 +5,41 @@ let
...
+  debugDeploy = let x = (builtins.getEnv "DEBUG_DEPLOY") != "";
+                in if !x then x else builtins.trace "DEBUG_DEPLOY is set, enriching deployment with debugging tools" x;
 
   # TODO: sanity check there's no duplicate nodes for same index
   # https://github.com/NixOS/nixops/blob/e2015bbfcbcf7594824755e39f838d7aab258b6e/nix/eval-machine-info.nix#L173
-  mergeNodes = mergeAttrs;
   mergeAttrs = nodes: lib.foldAttrs (a: b: a) [] nodes;
 
-  mkNodes = nodes: config: lib.mapAttrs (name: value: config value.i value.region) nodes;
+  mkNodesUsing = constructor: nodes: lib.mapAttrs  (name: nodeParams: constructor nodeParams) nodes;
```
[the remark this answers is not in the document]
deepfire: @domenkozar, any suggestions?  `config` was even worse, IMO..

modules/cardano-node-aws.nix
```diff
@@ -1,38 +1,51 @@
...
-      deployment.ec2.region = region;
+      deployment.ec2.region = mkForce params.region;
       deployment.ec2.accessKeyId = accessKeyId;
-      deployment.ec2.keyPair = resources.ec2KeyPairs.${keypairFor accessKeyId region};
-      deployment.keys = optionalAttrs (cfg.productionMode && !cfg.hasExplorer) {
-        "key${toString (nodeIndex + 1)}" = {
-          keyFile = ./. + "/../keys/key${toString (nodeIndex + 1)}.sk";
+      deployment.ec2.keyPair = resources.ec2KeyPairs.${keypairFor accessKeyId params.region};
+      deployment.ec2.securityGroups = mkForce (map sgByName params.sgNames);
+      deployment.keys = (optionalAttrs (cfg.productionMode  && params.type == "core" && !cfg.hasExplorer)
```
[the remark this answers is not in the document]
deepfire: right, that was caught&fixed in the `staging-federation` -- backporting : -)

modules/cardano-node-config.nix
```diff
@@ -1,17 +1,49 @@
...
+    sep                  = ", ";
+  in {
     imports = [
       ./common.nix
     ];
 
+    services.dnsmasq.enable = true;
+    services.dnsmasq.servers = [ "127.0.0.1" ];
+
+    networking.extraHosts =
```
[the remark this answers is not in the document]
deepfire: The kind of names that NixOS generates don't have a `.`, so aren't FQDN-like.

This breaks the `dns` library, which rejects them.

modules/hydra-master.nix
```diff
@@ -26,6 +27,8 @@ in {
     mode = "0440";
   };
 
+  environment.systemPackages = [ iohk-pkgs.iohk-ops ];
```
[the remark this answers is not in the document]
deepfire: It absolutely doesn't -- but previously you have added `iohk-ops` to _all_ machines.

I just made it so that it's not deployed on the cluster -- retaining other places, as per your original design.

Tell me if I should drop it from Hydra as well.

iohk/NixOps.hs
deepfire: I once tried to override `nixops`, but it doesn't seem like a normal derivation..  I guess I need someone's (@cleverca22 ?) help here.

cleverca22: one thought i had is to just use IFD, exactly the same way we do for stack2nix

accessKeyId.org
```diff
@@ -0,0 +1,70 @@
+* =cardano-nodes-config.nix=, the information hub
```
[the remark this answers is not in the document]
deepfire: Fixed.

deployments/cardano-explorer-config.nix
```diff
@@ -0,0 +1,21 @@
...
+let
+  config     = import ./cardano-nodes-config.nix { inherit accessKeyId deployerIP systemStart environment; };
+  spec       = {
+    region   = "eu-central-1";
+    type     = "other";
+    kademlia = false;
+    peers    = [];
+  };
+in rec {
+        inherit (spec) region type kademlia peers;
```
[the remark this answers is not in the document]
deepfire: Done.

deployments/cardano-explorer-config.nix
```diff
@@ -0,0 +1,21 @@
...
+        inherit (config.globalArgs) allNames environment firstRelayIndex nRelays systemStart;
+              i = length allNames; # assign last index
+     relayIndex = null;
+           name = "explorer";
+                  ## For the SG definitions look below in this file:
```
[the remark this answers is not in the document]
deepfire: Clarified : -)

deployments/cardano-nodes-env-production.nix
```diff
@@ -1,18 +1,20 @@
...
-  nodes = import ./cardano-nodes-config.nix { inherit nodeLimit; };
-  nodeProdConf = import ./../modules/cardano-node-prod.nix;
-in {
+  nodeArgs    = (import ./cardano-nodes-config.nix { inherit accessKeyId deployerIP systemStart environment; }).nodeArgs;
+  nodeEnvConf = import ./../modules/cardano-node-prod.nix;
+in
+{
   resources = {
-    elasticIPs = mkNodeIPs nodes accessKeyId;
+    elasticIPs = mkNodeIPs config.nodeArgs accessKeyId;
```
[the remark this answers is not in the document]
deepfire: Fixed..

deployments/firewalls.nix
```diff
@@ -0,0 +1,11 @@
+{ accessKeyId, deployerIP, systemStart, environment, ... }:
```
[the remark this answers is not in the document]
deepfire: Done.

iohk/CardanoCSL.hs
```diff
@@ -157,7 +137,7 @@ getSmartGenCmd o c = runError $ do
   dhtfile <- lift $ Prelude.readFile "static/dht.json"
   peers <- ExceptT $ return $ getPeers c config dhtfile nodes
   (_, sgIp) <- fmap T.strip <$> shellStrict ("curl " <> fromURL Ops.awsPublicIPURL) empty
-  let sgDhtKey = fromJust $ dhtfile ^? key "node100" . _String
+  -- let sgDhtKey = fromJust $ dhtfile ^? key "node100" . _String
```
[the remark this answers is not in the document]
deepfire: Done.

lib.nix
deepfire: I understand the idea is to always deploy dontStrip..

modules/cardano-node-development.nix
```diff
@@ -0,0 +1,16 @@
+with (import ./../lib.nix);
+
+{ config, pkgs, resources, ... }: {
+  imports = [
+    ./common.nix
```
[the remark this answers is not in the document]
deepfire: Done.

modules/cardano-node-development.nix
```diff
@@ -0,0 +1,16 @@
...
+  ];
+
+  # DEVOPS-64: disable log bursting
+  services.journald.rateLimitBurst = 0;
```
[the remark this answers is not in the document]
deepfire: Done.

modules/cardano-node.nix
```diff
@@ -98,6 +86,11 @@ in {
         description = "IP:Port of the EKG telemetry sink";
         default = null;
       };
+      serveEkg = mkOption {
+        type = types.bool;
+        description = "Serve EKG raw on port 80";
```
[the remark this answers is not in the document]
deepfire: Fixed.

shell.nix
```diff
@@ -2,31 +2,30 @@ let
   localLib = import ./lib.nix;
   nixpkgs  = localLib.fetchNixPkgs;
   pkgs     = import nixpkgs {};
-  iohkpkgs = import ./default.nix {};
+  hpkgs    = pkgs.haskell.packages.ghc802;
+  hpkgs'   =
+  hpkgs.override {
```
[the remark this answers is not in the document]
deepfire: Removed.

iohk/NixOps.hs
deepfire: Fixed.

config-system-start.nix
deepfire: Dropped.

deployments/cardano-nodes-config.nix
deepfire: Got annoyed by these traces as well, so dropped them
.

lib.nix
domenkozar: Whole linux ecosystem is written in C, that's hardly an argument. Besides, if someone gets access to machines - the security is long gone.

lib.nix
domenkozar: This doesn't work anyway, we'll add debug flag later on

modules/amazon-base.nix
```diff
@@ -1,13 +1,12 @@
...
 with (import ./../lib.nix);
 
+{ name, config, resources, ... }:
+
 {
  deployment.targetEnv = "ec2";
  deployment.ec2.instanceType = mkDefault "t2.large";
- deployment.ec2.region = mkDefault region;
- deployment.ec2.keyPair = mkDefault (resources.ec2KeyPairs.${keypairFor config.deployment.ec2.accessKeyId region});
- deployment.ec2.securityGroups = mkDefault [resources.ec2SecurityGroups."allow-all-${config.deployment.ec2.region}" ];
```
domenkozar: Removing this breaks Hydra and infrastructure in general, I think.

deepfire: Thanks, I'll look into it!

deepfire:
@domenkozar, I've reverted https://github.com/input-output-hk/iohk-ops/pull/130/commits/3ae78f32f07a489b38c365c4c767bcd7d3974b99, because it makes Hydra evaluation fail in odd circumstances (i.e. not always):

```
Errors occurred at 2017-09-07 00:36:09.

in job ‘cardano-sl-tools-static’:
attribute ‘cardano-sl-tools-static’ missing, at [1m/nix/store/5p152d0414swc9bzfkvr6m7mg3hijaaa-git-export/jobsets/cardano.nix[0m:13:21
```

deepfire:
@domenkozar, I've also had to fix https://github.com/input-output-hk/iohk-ops/pull/130/commits/6e0c5f520ae712f67be5641b7a3eb89e38d8fb07, because it had a regression.

domenkozar:
https://github.com/input-output-hk/iohk-ops/commit/3ae78f32f07a489b38c365c4c767bcd7d3974b99 needs https://github.com/input-output-hk/iohk-ops/pull/134 to be merged to pass CI

deepfire:
@domenkozar, CI is green, as of https://github.com/input-output-hk/iohk-ops/pull/130/commits/705ea9c8bc7d22e3af420f05909d8399b46f77bf

scripts/travis.sh
```diff
@@ -15,12 +15,12 @@ WITH_REPORT_SERVER=${8:-true}
 WITH_INFRA=${9:-true}
 
 # 0. Check all scripts compile
-nixops --version
-nix-shell --run "./scripts/aws.hs --help"
-${IOHK_OPS} --help
+# nixops --version
```
domenkozar: This shouldn't be commited ever, CI check should stay in place

scripts/travis.sh
```diff
@@ -0,0 +1,73 @@
...
+set -xeu
+
+NIXOPS=$(nix-build -A nixops)/bin/nixops
+
+IOHK_OPS=${1:-$(nix-build -A iohk-ops)/bin/iohk-ops}
+CLEANUP_DEPLOYS=${2:-true}
+CLEANUP_CONFIGS=${3:-true}
+WITH_STAGING=${4:-true}
+WITH_PRODUCTION=${5:-true}
+WITH_DEVELOPMENT=${6:-}         ## XXX: stop-gap fix for CI
```
domenkozar: This needs to be tested by CI

deployments/infrastructure.nix
```diff
@@ -5,8 +5,9 @@ let
   iohk-pkgs = import ../default.nix {};
   mkHydraBuildSlave = { config, pkgs, ... }: {
     imports = [
-      ./../modules/hydra-slave.nix
+      ./../modules/amazon-base.nix
```
domenkozar: This is wrong in general, previously we had `*-aws-*.nix` that added aws backend specific configuration, so one could use in theory a different backend. Such changes will make it harder to test things on virtualbox or us moving to a provider like GCE.

domenkozar: I wonder what this commit really fixes? It should be reverted.

modules/report-server.nix
```diff
@@ -31,15 +31,18 @@ in {
 
   config = {
 
-    # TODO: remove
-    global.organisation = params.org;
-    global.dnsHostname = mkForce "report-server";
+    global = {
```
domenkozar: The TODO should be left in tact, we are mixing nixops options together with a module that describes how to deploy it.

modules/report-server.nix
```diff
@@ -31,15 +31,18 @@ in {
...
+      dnsHostname              = mkForce "report-server";
+    };
+
     deployment.ec2.region         = mkForce params.region;
     deployment.ec2.accessKeyId    = params.accessKeyId;
     deployment.ec2.keyPair        = resources.ec2KeyPairs.${params.keyPairName};
-    deployment.ec2.securityGroups = map (resolveSGName resources) [
-      "allow-to-report-server-${config.deployment.ec2.region}"
-    ];
+    deployment.ec2.securityGroups = optionals (! config.global.omitDetailedSecurityGroups)
```
domenkozar: I don't think this will work, it needs a security group otherwise you can't connect via SSH.

Moreover, we should set different SG per env - as it is deployment specific

staging-testnet.yaml
```diff
@@ -29,5 +29,5 @@ elements:
 - Nodes
 - Explorer
 - ReportServer
-nixops: /nix/store/mb4rfbkliwjvcdcnqzgg4bgi45i8pb3s-nixops-1.6pre0_abcdef/bin/nixops
+nixops: /nix/store/7ckasdkad0vl5y1lpi2arlh6fipfmkv7-nixops-1.6pre0_abcdef/bin/nixops
```
domenkozar: This should have a corresponding change in default.nix

modules/cardano-service.nix
```diff
@@ -45,7 +45,7 @@ let
     (optionalString (!cfg.enableP2P) "--kademlia-explicit-initial --disable-propagation ${smartGenPeer}")
     "--topology ${cfg.topologyYaml}"
     "--node-id ${params.name}"
-    "--config-name testnet_staging_full"
+    "--custom-config-name testnet_staging_full"
```
domenkozar: I wonder, is this required to be set?

deployments/infrastructure.nix
deepfire: @domenkozar, I think we previously discussed and agreed on the `-aws-` aspect separation being too much of a _de-facto_ burden for the benefits it currently provides (none).

If we want to reverse that decision, that's okay, I guess -- let's just be explicit about it, since a lot of the current code assumes AWS.

modules/report-server.nix
deepfire: I agree, yes.

modules/report-server.nix
deepfire: The `omitDetailedSecurityGroups` option is a network-wide option to basically use `allow-all-*`.

It allows the two SG deployment variants to co-exist.

staging-testnet.yaml
deepfire: I think that with `nixops` being specified in `default.nix`, we shouldn't need `nixops` in our deployment config anymore.

But that would take some work to implement -- should I add a checkbox in the PR merge checklist, @domenkozar?

deepfire: Also, this particular change is a revert of  _previous_ errorneous change -- apparently Nixops build isn't really deterministic, so on my system it yields a different store path -- with the same Nixpkgs.

If that even makes sense..

modules/cardano-service.nix
deepfire: Yes, this night's deployment failed because of:

1. Duncan needed a new topo file
2. The Cardano changes required to accomodate it were in `master`
3. `master` now takes this flag to choose the genesis it starts with -- precisely the direction we want, but in this case for me it meant..
4. Having to wait for the morning, until George tells me about the config changes.

deepfire: All of this is in the Slack archives of #devops and #sl..

deepfire: Because we've communally agreed we need to go back, and re-deploy the old staging, this needs to be reverted, of course (old Cardano won't accept this option).

deployments/infrastructure.nix
deepfire: This commit moves development-agnostic things into commonly included locations.

deepfire: Correction: deployment-agnostic things.

domenkozar: I have never agreed on this, we should keep in mind to keep deployments agnostic to nixops backend. That is why `-aws` files existed - I don't understand why these changes were reverted.

staging-testnet.yaml
domenkozar: That shouldn't happen if same nixpkgs commit is used.

modules/hydra-master.nix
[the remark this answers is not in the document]
deepfire: Dropped.

modules/hydra-slave.nix
```diff
@@ -1,12 +1,17 @@
...
+  iohk-pkgs = import ../default.nix {};
+in
 {
   nix.gc = {
     automatic = true;
     dates = "05:15";
     options = ''--max-freed "$((32 * 1024**3 - 1024 * $(df -P -k /nix/store | tail -n 1 | ${pkgs.gawk}/bin/awk '{ print $4 }')))"'';
   };
 
+  environment.systemPackages = [ iohk-pkgs.iohk-ops ];
```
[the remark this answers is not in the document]
deepfire: Dropped.

staging-testnet.yaml
deepfire: It shouldn't, yet it does.. : -/  Some debugging is in order..

deployments/infrastructure.nix
deepfire: Well, my mistake for understanding it like this, then.

This was quite a bit ago -- during a standup, where I've complained that implementing the core/relay split is getting a bit too complicated with the target aspect separation, and that there already are places (not introduced by me) where the target (physical spec) isolation was not getting respected.  You then agreed that the big change we've been implementing soft of conflicted with the physical spec (aka target) separation.

I understood this as license to inline AWS into common parts.

deepfire: I guess the question now is what is our plan for moving forward with the physical spec aspec separation.

domenkozar: I've agreed that we don't have time to fix this now, but that doesn't mean we should make the situation worse.

domenkozar:
@deepfire https://github.com/input-output-hk/iohk-ops/pull/130/commits/2a0886598e0a523f71052997b259c9470865a18f and https://github.com/input-output-hk/iohk-ops/pull/130/commits/a7aeeab37518debf807c4f1b99283673ae898e04 and https://github.com/input-output-hk/iohk-ops/pull/130/commits/391fa8c7ef25df91df8d09f29c3dee717919a5c1 and https://github.com/input-output-hk/iohk-ops/pull/130/commits/5ac77bdee0238a630f52d5225fdd35721573e287 can be reverted by just doing:

` configurationYaml = ${ ./configuration.mainnet.yaml }`

If we'll need to use different configs I suggest we `dconfig` in the filename using: `configurationYaml = ${ ./configuration. + dconfig + ".yaml" }`

default.nix
```diff
@@ -19,7 +19,6 @@ let
   cardano-sl-src = builtins.fromJSON (builtins.readFile ./cardano-sl-src.json);
   cardano-sl-pkgs = import (pkgs.fetchgit cardano-sl-src) {
                       gitrev = cardano-sl-src.rev;
-                      inherit dconfig;
```
domenkozar: You should have merged https://github.com/input-output-hk/iohk-ops/pull/146 to fix this - I'll do it now

deployments/cardano-nodes-env-development.nix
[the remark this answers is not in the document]
deepfire: Fixed.

lib.nix
deepfire: The attack scenario is that the attacker is outside, indeed, and is producing traffic that causes a legitimate use of `tcpdump` to give attacker a shell.

modules/hydra-master.nix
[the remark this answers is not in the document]
deepfire: The question was more complicated, apparently, as we've reversed our decision here once more.

modules/amazon-base.nix
deepfire: Fixed.

iohk/NixOps.hs
deepfire: Right, that's how it ended up being implemented.

deployments/infrastructure-target-aws.nix
```diff
@@ -29,17 +30,46 @@ rec {
...
+      };
+      "allow-public-www-${region}-${org}" = {
+        inherit region accessKeyId;
+        description = "WWW";
+        rules = [{
+          protocol = "tcp"; # TCP
+          fromPort = 443; toPort = 443;
+          sourceIp = "0.0.0.0/0";
+        }];
+      };
```
deepfire: №3

We probably want to have the more centralised configuration approach for security groups applied to infrastructure as well.

deployments/infrastructure.nix
```diff
@@ -41,15 +43,6 @@ in {
...
 
-      users.production = {
-        description     = "cardano production";
-        group           = "production";
-        createHome      = true;
-        isNormalUser = true;
-        openssh.authorizedKeys.keys = [];  # this account is obsolete
-      };
-      groups.production = {};
-
```
deepfire: №4

Do we want to rename `live-production` back to `production`?

deployments/report-server-target-aws.nix
```diff
@@ -1,14 +0,0 @@
...
-  network.description = "Cardano SL";
-
-  report-server = { config, pkgs, resources, ... }: {
-    imports = [
-      ./../modules/amazon-base.nix
-    ];
-
-    deployment.ec2.accessKeyId = accessKeyId;
-  };
-}
```
deepfire: №5

We probably want to restore the physical machine spec (AWS, GCE, ...) aspect separation.

deployments/security-groups.nix
```diff
@@ -0,0 +1,133 @@
...
+          (fold (x: y: x // y) {}
+          (         (centralRegionSGs  centralRegion)
+
+           ++  map  (regionSGs         { nodePort = 3000; }) # TODO: 'config' is mostly empty here.
+                    globals.allRegions
+
+           ++  map  (orgXRegionSGs)
+                    globals.orgXRegions
+
+           ++  map  (coreSGs           3000 resources.elasticIPs) # TODO: config for port
```
deepfire: №6

We should pass through the actual node's listen port here.

deployments/security-groups.nix
```diff
@@ -0,0 +1,133 @@
...
+        regionSGs      = { nodePort }: region: {
+            "allow-kademlia-public-udp-${region}" = {
+              inherit region accessKeyId;
+              description = "Kademlia UDP public";
+              rules = [{
+                protocol = "udp"; # UDP
+                sourceIp = "0.0.0.0/0";
+                fromPort = 1024; toPort = 65535;
+              }];
+            };
```
deepfire: №6
There should be a less verbose way to define those.

globals.nix
```diff
@@ -0,0 +1,98 @@
...
+                       { name  = "report-server";
+                         value = { org      = defaultOrg;
+                                   region   = centralRegion;
+                                   zone     = centralZone;
+                                   type     = "other";
+                                   kademlia = false;
+                                   peers    = [];
+                                   address  = "report-server.cardano";
+                                   port     = null; }; };
+
```
deepfire: №7

There is a tension between:
1. having simpler definitions for non-node machines
2. having a more uniform definition mechanism for all machines

I'm not sure this tension cannot be resolved at all -- we should be able to have both, with careful planning.

.travis.yml
```diff
@@ -1,39 +1,15 @@
...
-  - nix-env -f '<nixpkgs>' -i -A nixopsUnstable -A git
-  - nix-env -iA iohk-ops       -f 'default.nix'
-  - touch static/datadog-{api,application}.secret
-  - echo "secret" > static/tarsnap-cardano-deployer.secret
-  - mkdir keys
-  - touch keys/key{0,1,2,3,4,5,6,7,8,9,10,11,12,13,41}.sk
+  - nix-env -f '<nixpkgs>'   -iA git
 script:
   - git --version
   - ./scripts/find-all-revisions.sh
```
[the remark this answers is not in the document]
deepfire: Done -- Domen has addressed this as part of https://github.com/input-output-hk/iohk-ops/pull/176

deployments/report-server-target-aws.nix
deepfire: Filed as https://issues.serokell.io/issue/DEVOPS-456

deployments/cardano-explorer-env-production.nix
```diff
@@ -1,19 +1,21 @@
...
+    datadogMonitors = (with (import ./../modules/datadog-monitors.nix); {
+      cardano_explorer_process = mkMonitor (cardano_explorer_process_monitor // {
+        message = pagerDutyPolicy.nonCritical;
+        monitorOptions.thresholds = {
+          warning = 3;
+          critical = 4;
+          ok = 1;
+        };
+      });
+    });
```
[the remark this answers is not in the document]
domenkozar: Yes, let's have an issue to refactor datadog monitors abstraction.

deployments/infrastructure.nix
domenkozar: Nah, let's leave it as is. Not worth the time.

deployments/security-groups.nix
deepfire: Filed as https://issues.serokell.io/issue/DEVOPS-457

globals.nix
domenkozar: This will be a result of DEVOPS-445

deployments/cardano-explorer-env-production.nix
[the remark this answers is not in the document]
deepfire: Filed https://issues.serokell.io/issue/DEVOPS-459

iohk/Topology.hs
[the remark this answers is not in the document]
deepfire: Now, that we're probably going to merge things into topology, we're got an even stronger case for having this separate implementation.

tests/simple-node.nix
```diff
@@ -1,12 +1,15 @@
+let
+  nodeMap = import ./cardano-node-simple-config.nix;
+in
 import <nixpkgs/nixos/tests/make-test.nix> ({ pkgs, ... }: {
   name = "simple-node";
   nodes = {
     machine = { config, pkgs, ... }: {
-      imports = [ (import ../modules/cardano-node-config.nix 0 "") ];
+      imports = [ (import ../modules/cardano.nix (nodeMap.machine)) ];
```
deepfire: https://issues.serokell.io/issue/DEVOPS-449: eliminate `call-on-imports`

tests/simple-node-nixops.nix
```diff
@@ -1,11 +1,14 @@
+let
+  nodeMap = import ./cardano-node-simple-config.nix;
+in
 import <nixpkgs/nixos/tests/make-test.nix> ({ pkgs, ... }: {
   name = "boot";
   nodes = {
     machine = { config, pkgs, ... }: {
-      imports = [ (import ../modules/cardano-node-config.nix 0 "") <nixops/nix/options.nix> <nixops/nix/resource.nix> ];
+      imports = [ (import ../modules/cardano.nix (nodeMap.machine)) <nixops/nix/options.nix> <nixops/nix/resource.nix> ];
```
deepfire: https://issues.serokell.io/issue/DEVOPS-449: eliminate `call-on-imports`

modules/report-server.nix
```diff
@@ -1,41 +1,73 @@
...
+        default = (import ./../default.nix {}).cardano-report-server-static;
+      };
     };
   };
 
-  imports = [ ./common.nix ];
-
   config = {
+
+    # TODO: remove
```
deepfire: Not really remove, but rather reconsider as part of https://issues.serokell.io/issue/DEVOPS-450

The `organisation` and `dnsHostname` are useful abstractions that are beneficial across the entire deployment.

modules/network-wide.nix
```diff
@@ -0,0 +1,86 @@
...
+                             config.global.dnsDomainname != null) {
+      accessKeyId = config.deployment.ec2.accessKeyId;
+      hostName    = config.global.dnsHostname +"."+ config.global.dnsDomainname;
+    };
+  };
+  options = {
+    global = mkOption {
+      description = "IOHK global option group.";
+      default = {};
+      type = with types; submodule {
```
deepfire: Some of those options are unused / not-really-useful, in that they are not truly network-wide, so cannot inform resource definitions.

topology-mainnet-infra.yaml
```diff
@@ -0,0 +1,8 @@
+nodes:
```
domenkozar: This shouldn't be needed :)

topology-min.yaml
```diff
@@ -0,0 +1,15 @@
+nodes:
```
domenkozar: @deepfire do we still need this?

modules/datadog-monitors.nix
```diff
@@ -79,20 +88,74 @@ rec {
...
+  cardano_explorer_process_monitor = {
+    name = "cardano-explorer process is down";
+    type = "service check";
+    query = config: "\"process.up\".over(\"env:${config.deployment.arguments.environment},depl:${config.deployment.name}\",\"process:cardano-explorer\").by(\"host\",\"process\").last(6).count_by_status()";
+    monitorOptions.thresholds = {
+      warning = 4;
+      critical = 5;
+      ok = 2;
+    };
+  };
```
deepfire: Again, https://issues.serokell.io/issue/DEVOPS-459

modules/cardano.nix
```diff
@@ -0,0 +1,114 @@
...
+      let sgNames =
+           optionals typeIsExplorer              [ "allow-to-explorer-${params.region}" ]
+        ++ optionals typeIsCore                  [ "allow-cardano-static-peers-${params.name}-${params.region}-${params.org}" ]
+        ++ optionals typeIsRelay                 [ "allow-kademlia-public-udp-${params.region}"
+                                                   "allow-cardano-public-tcp-${params.region}" ]
+        ++ optionals config.global.enableEkgWeb  [ "allow-ekg-public-tcp-${params.region}-${params.org}" ];
+      in map (resolveSGName resources)
+             (if config.global.omitDetailedSecurityGroups
+              then [ "allow-all-${params.region}-${params.org}" ]
+              else sgNames);
```
deepfire: This implementation should be reconsidered.  Perhaps as part of https://issues.serokell.io/issue/DEVOPS-460

We don't really want environment/component sensitivity hard-coded here.

modules/cardano-service.nix
```diff
@@ -1,31 +1,20 @@
-{ config, pkgs, lib, options, ... }:
-
 with (import ./../lib.nix);
 
+globals: params:
```
deepfire: Again, https://issues.serokell.io/issue/DEVOPS-449

modules/cardano-production.nix
```diff
@@ -0,0 +1,15 @@
...
+
+params:
+{ name, config, pkgs, resources, ... }: {
+  imports = [
+    ./production.nix
+    ./datadog.nix
+    ./papertrail.nix
+  ];
+
+  global.dnsHostname = if params.typeIsRelay then "cardano-node-${toString params.relayIndex}" else null;
```
deepfire: It _does_ look like `typeIsRelay` is at least a good match here.

What are other possible good ways to state the same thing?

modules/cardano-explorer.nix
```diff
@@ -1,54 +1,48 @@
-let
-
-  explorer = (import ./../default.nix {}).cardano-sl-explorer-static;
-
-in
+with (import ./../lib.nix);
 
+globals: params:
```
deepfire: Again, https://issues.serokell.io/issue/DEVOPS-449

modules/cardano-development.nix
```diff
@@ -0,0 +1,8 @@
+with (import ./../lib.nix);
+
+params:
```
deepfire: Again, https://issues.serokell.io/issue/DEVOPS-449

lib.nix
```diff
@@ -2,37 +2,37 @@
...
+    };
 
-  # Given a list of integers and a function,
-  # generate an attribute set with that many nodes and call the function with node index
-  genNodes = nodeLimit: ids: f:
-             genAttrs' (lib.flip builtins.filter ids
-                          (id: id < nodeLimit))
-                       (i: "node${toString i}") f;
+  centralRegion = "eu-central-1";
+  centralZone   = "eu-central-1b";
```
deepfire: I'm not _really_ sure these definitions belong here..

Moved them here out of desperation, in attempt to pick apart the infamous `globals.nix`.

lib.nix
```diff
@@ -2,37 +2,37 @@
...
+  lib = pkgs.lib;
 in lib // (rec {
-  # Allows to also generate the key compared to upstream genAttrs
-  genAttrs' = names: fkey: fname:
-    lib.listToAttrs (map (n: lib.nameValuePair (fkey n) (fname n)) names);
+  ## nodeElasticIP :: Node -> EIP
+  nodeElasticIP = node:
+    { name = "${node.name}-ip";
+      value = { inherit (node) region accessKeyId; };
+    };
```
deepfire: Similarly, not sure this belongs here..

PR #133, opened 2017-08-30
[DEVOPS-236] Staging federation

deepfire:
This allows for three-AWS-account deployment of `staging` environment.

Review checklist:

- [ ] have `staging` PR merged
- [ ] resurrect development & production environments
- [ ] ensure this review checklist is complete


model.nix
```diff
@@ -0,0 +1,53 @@
+{ ... }:
```
domenkozar: This shouldn't be committed

modules/amazon-base.nix
```diff
@@ -1,12 +1,13 @@
-{ name, config, resources, ... }:
-
 with (import ./../lib.nix);
 
+globals: params:
```
domenkozar: Why is this better? We want to use more of NixOS module system, not less.

deepfire:
It was decided to merge this branch into `staging`, instead.

PR #135, opened 2017-09-08
Build nixops with Nix and reference it where needed

domenkozar:


default.nix
```diff
@@ -20,6 +20,16 @@ let
...
+  nixops = 
+    let
+      # nixopsUnstable = /path/to/local/src
+      nixopsUnstable = pkgs.fetchFromGitHub {
+        owner = "NixOS";
+        repo = "nixops";
+        rev = "c06c0e79ab8d7a58d80b1c38b7ae4ed1a04322f0";
+        sha256 = "1fly6ry7ksj7v5rl27jg5mnxdbjwn40kk47gplyvslpvijk65m4q";
+      };
+    in (import "${nixopsUnstable}/release.nix" {}).build.x86_64-linux;
```
deepfire: Aha, import-from-derivation!

@domenkozar, that was the part I was missing : -)

PR #141, opened 2017-09-15
--evaluate-only -> --dry-run

domenkozar:
Sadly `--evaluate-only` doesn't evaluate each machine, so we have to use `--dry-run` for this. Waiting on CI to confirm.


domenkozar:
@deepfire I'm going to merge this in staging since it works as intended, but CI now fails since evaluation for explorer is broken.

PR #143, opened 2017-09-18
[DEVOPS-338] More complete travis sh

deepfire:
https://issues.serokell.io/issue/DEVOPS-338

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,8 @@ smart-gen-*
 static/id_buildfarm
 *prof
 keys/*.key
+/test-*.yaml
 static/*.secret
+static/github_token
 pkgs/stack2nix
 scripts/iohk-nixpkgs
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -21,6 +21,8 @@ smart-gen-*
 static/id_buildfarm
 *prof
 keys/*.key
+/test-*.yaml
```
domenkozar: What's this about?

domenkozar: Aren't those always deleted by `cleanup` function?

deepfire: They are, but it still causes a turmoil in your git status while the test is running.. : -)

domenkozar: Ideally then we'd use `mktemp -d` for intermediate artifacts and cleanup the directory after everything is done

deepfire: I agree, this could be the next improvement we make.

domenkozar:
@deepfire CI fails

domenkozar:
@deepfire this hasn't passed CI yet, why did you merge it?

deepfire:
@domenkozar, it was an unfortunate accident -- I've had to have this locally, because otherwise `static/github_token` was in the `git status`.

deepfire:
@domenkozar, but I think there is no reason for worry:
1. if it doesn't work I'm fixing it as I go
2. Travis doesn't build much of anything today, anyway : -/

domenkozar:
How is that a blocker or a real problem? :)

domenkozar:
I'd like for workflows to be respected, things must pass CI

deepfire:
@domenkozar, I totally agree -- it was an unfortunate accident, as I said.

domenkozar:
If things are accidentally merged, they should be reverted.

deepfire:
@domenkozar, if you think it would be efficient use of our time to do this -- of course, I can do that.

domenkozar:
Yes, I think respecting the workflows is an efficient use of time.

PR #148, opened 2017-09-23
Deploying deployer

deepfire:
Merge checklist:

- [ ] allow deployment of both infras in parallel
- [ ] ensure that this merge checklist is satisfactorily complete

### DIFF
--- a/deployments/infrastructure-target-aws.nix
+++ b/deployments/infrastructure-target-aws.nix
@@ -17,7 +17,7 @@ in rec {
       ebsInitialRootDiskSize = mkForce 200;
       associatePublicIpAddress = true;
       securityGroups = [
-        resources.ec2SecurityGroups."allow-all-ssh-${region}-${org}"
+        resources.ec2SecurityGroups."allow-deployer-ssh-${region}-${org}"
         resources.ec2SecurityGroups."allow-public-www-${region}-${org}"
       ];
     };
--- a/deployments/infrastructure-target-aws.nix
+++ b/deployments/infrastructure-target-aws.nix
@@ -1,19 +1,25 @@
-{ IOHKaccessKeyId, ... }:
+{ deployerIP, IOHKaccessKeyId, ... }:
 
 with (import ./../lib.nix);
-rec {
+let org = "IOHK";
+    region = "eu-central-1";
+    accessKeyId = IOHKaccessKeyId;
+in rec {
   hydra = { config, pkgs, resources, name, ... }: {
-
     imports = [
       ./../modules/amazon-base.nix
     ];
 
     deployment.ec2 = {
-      accessKeyId = IOHKaccessKeyId;
-
+      inherit accessKeyId;
+      elasticIPv4 = resources.elasticIPs.hydra-ip;
       instanceType = mkForce "r3.2xlarge";
       ebsInitialRootDiskSize = mkForce 200;
       associatePublicIpAddress = true;
+      securityGroups = [
+        resources.ec2SecurityGroups."allow-all-ssh-${region}-${org}"
+        resources.ec2SecurityGroups."allow-public-www-${region}-${org}"
+      ];
     };
     deployment.route53.accessKeyId = config.deployment.ec2.accessKeyId;
     deployment.route53.hostName = "${name}.aws.iohkdev.io";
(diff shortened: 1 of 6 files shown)


topology-mainnet-infra.yaml
```diff
@@ -0,0 +1,8 @@
+nodes:
```
domenkozar: will this deploy something or is it just for a stub?

deepfire: @domenkozar, no, just a stub, right.

deployments/infrastructure-target-aws.nix
```diff
@@ -33,11 +39,54 @@ rec {
...
       associatePublicIpAddress = true;
+      securityGroups = [
+        resources.ec2SecurityGroups."allow-all-ssh-${region}-${org}"
+      ];
+    };
+  };
+
+  resources = {
+    elasticIPs = {
+      hydra-ip    = { inherit region accessKeyId; };
```
domenkozar: These should still be in production config

deployments/infrastructure-target-aws.nix
```diff
@@ -33,11 +39,54 @@ rec {
...
+      cardanod-ip = { inherit region accessKeyId; };
+    };
+    ec2SecurityGroups = {
+      "allow-deployer-ssh-${region}-${org}" = {
```
domenkozar: This security group is not needed because `allow-all-ssh` overrides it to all IPs

deployments/infrastructure-target-aws.nix
```diff
@@ -33,11 +39,54 @@ rec {
...
+        inherit region accessKeyId;
+        description = "WWW";
+        rules = [{
+          protocol = "tcp"; # TCP
+          fromPort = 443; toPort = 443;
+          sourceIp = "0.0.0.0/0";
+        }];
+      };
     };
+    datadogMonitors = (with (import ./../modules/datadog-monitors.nix); {
```
domenkozar: This should be in production config only

topology-mainnet-infra.yaml
domenkozar: let's add a comment here to explain why this file exists for future refactoring.

deployments/infrastructure-target-aws.nix
deepfire: ..but it would break the AWS separation!

deployments/infrastructure-target-aws.nix
deepfire: We need to only allow the rest of infra to be accessed through the deployer only?

It was my intention, but I didn't carry it through..

deployments/infrastructure-target-aws.nix
deepfire: Yeah, that slipped..

deepfire: Fixed that, and also moved tarsnap defns to the common `infrastructure.nix`.

deployments/infrastructure-target-aws.nix
domenkozar: I'm not following, production is only on AWS. There won't be any other productions :)

deployments/infrastructure-target-aws.nix
domenkozar: Then let's add a comment ` TODO: allow hydra and slaves to only have SSH access via deployer`

deepfire: @domenkozar, already implemented!

domenkozar: Ah I see now :)

deployments/infrastructure.nix
```diff
@@ -63,6 +51,11 @@ in {
       groups.live-production = {};
     };
 
+    deployment.keys.tarsnap = {
```
domenkozar: This should be set only in production

deepfire: Done

deepfire: (I didn't expect us to not want to backup the current, non-`mainnet` deployer..)

deployments/infrastructure-target-aws.nix
deepfire: So why having AWS separate at all?

deepfire: That's part of my confusion, actually, @domenkozar..  you used to say that we want to keep AWS-specific configuration (aka physical spec) separate.

domenkozar: Yes, it's for deploying testing changes for AWS.

deepfire: ..and this means the consistent way would be to have `infrastructure-env-production-target-aws.nix`

domenkozar: That would imply there are other (can exist) production deployment backends, which is false.

deepfire: @domenkozar, then why isn't the very same logic applicable to the cluster itself?

Please, don't get alarmed by my questions -- they are the root reason why I started folding `*-target-aws.nix` -- I'm not understanding the difference between approaches here -- and I thought we wanted a common approach..

deepfire: One possibility is that we want to be potentially able to distribute the cluster to GCE/Azure/whatever.

That I can understand.

But from this _might_ follow a requirement that we keep a deployer there as well -- at least for performance reasons.

Hence a non-AWS deployer could potentially make sense.

PR #151, opened 2017-09-24
ergo | iohk-ops get-journals:  replace ':' -> '_', to avoid requiring  --force-local with 'tar'

deepfire:


domenkozar:
Can you provide more context what this fixes?

deepfire:
@domenkozar, `:` are treated specially by tar, which requires the archive user to use the flag I mentioned -- which is a hindrance..

PR #152, opened 2017-09-25
[DEVOPS-378] setenv

deepfire:
`--build-only` was preventing `GC_INITIAL_HEAL_SIZE` from being set.

### DIFF
--- a/iohk/NixOps.hs
+++ b/iohk/NixOps.hs
@@ -841,7 +841,7 @@ deploy o@Options{..} c@NixopsConfig{..} dryrun buonly check reExplorer bumpSyste
     deployerIP <- establishDeployerIP o oDeployerIP
     setenv o "SMART_GEN_IP" $ getIP deployerIP
   when (elem Nodes cElements) $
-    setenv o "GC_INITIAL_HEAP_SIZE" (showT $ 8 * 1024*1024*1024) -- for 100 nodes it eats 12GB of ram *and* needs a bigger heap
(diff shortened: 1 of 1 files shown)


iohk/NixOps.hs
```diff
@@ -832,9 +839,9 @@ deploy o@Options{..} c@NixopsConfig{..} dryrun buonly check reExplorer bumpSyste
     cmd o "scripts/generate-explorer-frontend.sh" []
   when (dryrun /= DryRun && buonly /= BuildOnly) $ do
     deployerIP <- establishDeployerIP o oDeployerIP
-    export "SMART_GEN_IP" $ getIP deployerIP
-    when (elem Nodes cElements) $ do
-      export "GC_INITIAL_HEAP_SIZE" (showT $ 8 * 1024*1024*1024) -- for 100 nodes it eats 12GB of ram *and* needs a bigger heap
+    setenv o "SMART_GEN_IP" $ getIP deployerIP
+  when (elem Nodes cElements) $
+    setenv o "GC_INITIAL_HEAP_SIZE" (showT $ 8 * 1024*1024*1024) -- for 100 nodes it eats 12GB of ram *and* needs a bigger heap
```
domenkozar: This should be 15 GB now

deepfire: Ack

deepfire: @domenkozar, would you prefer that to be a parameter?

PR #155, opened 2017-09-29
[DEVOPS-385, 394] Monitor light fixes

deepfire:


modules/staging.nix
```diff
@@ -11,7 +11,7 @@ with import ./../lib.nix;
     };
 
     services = {
-      dd-agent.tags              = ["env:rc-staging"];
+      dd-agent.tags              = ["env:staging"];
```
jmitchell: At the moment it's useful having both `rc-staging` and `staging` tags in DD. I'd like to keep it that way, unless we will kill the existing `staging` deployment before deploying this change to `rc-staging`.

scripts/travis.sh
```diff
@@ -8,7 +8,7 @@ IOHK_OPS=${1:-$(nix-build -A iohk-ops)/bin/iohk-ops}
 CLEANUP_DEPLOYS=${2:-true}
 CLEANUP_CONFIGS=${3:-true}
 WITH_STAGING=${4:-true}
-WITH_PRODUCTION=${5:-true}
+WITH_PRODUCTION=${5:-}
```
jmitchell: Later we branch on whether `$WITH_PRODUCTION` is an empty string or not. Is the only way to pass arguments 6, 7, and 8 while keeping `$WITH_PRODUCTION` empty using a `""` placeholder? Is that what we want? I see `.travis.yml` doesn't pass any arguments, so it's using the default values, in this case an empty string.

modules/cardano.nix
```diff
@@ -57,6 +57,26 @@ globals: imports: params:
...
+    else ''
+    - name: cardano-node-simple
+      search_string: ['cardano-node-simple']
+      '') + ''
+    # nix string indentation hack (see DEVOPS-341)
+      exact_match: True
+      thresholds:
+        critical: [1, 1]
+    '';
+
```
jmitchell: Was this copied in from another branch? Were there any changes made to from the original?

I recall working on this code to add the indentation hack. I know the indentation hack was verified in some environment like staging or TN06.

jmitchell: Oh, right, it's being moved from `modules/cardano-staging.nix` with two changes:

* `exact_match` is enabled
* `cardano-node-simple` instead of `cardano-node` for the search string

scripts/travis.sh
deepfire: It is, correct.

deployments/infrastructure-env-production.nix
```diff
@@ -13,7 +13,7 @@ in {
       ./../modules/datadog.nix
     ];
 
-    services.dd-agent.tags = ["env:production"];
+    services.dd-agent.tags = ["env:production", "depl:${config.deployment.name}"];
```
jmitchell: It would be nice to have a tag that distinguishes mainnet nodes from production-grade infrastructure nodes. Perhaps `env:production-infra` would be good.

scripts/travis.sh
deepfire: And my taste tells me it's better, when it's parametrisable : -)

..esp when you do local testing iterations, where you want to eval different parts.

deployments/infrastructure-env-production.nix
jmitchell: If we do that, we'd need to be sure all the declaratively specified DD monitors are created for both envs.

modules/staging.nix
deepfire: fixed, we now have a proper `depl:` tag

deployments/cardano-nodes-env-staging.nix
```diff
@@ -7,12 +7,12 @@ let nodeMap = globals.nodeMap; in
...
   network.description = "Cardano Staging";
 
-  resources = {
+  resources = { config }: {
     elasticIPs = nodesElasticIPs nodeMap;
     datadogMonitors = (with (import ./../modules/datadog-monitors.nix); {
       cpu = mkMonitor (cpu_monitor // {
         message = pagerDutyPolicy.nonCritical;
-        query = config: "avg(last_5m):avg:system.load.norm.1{env:${config.deployment.name}} by {host} > 0.99";
+        query = config: "avg(last_5m):avg:system.load.norm.1{env:${environment},depl:${config.deployment.name}} by {host} > 0.99";
```
jmitchell: Is the intent to deploy these changes to staging for verification and then, if it works, go back and apply the equivalent changes to `deployments/cardano-nodes-env-production.nix`?

deployments/infrastructure-env-production.nix
deepfire: @jmitchell, we can do select those by intersectingf:
- `env: production` 
- `depl: infra`

scripts/travis.sh
jmitchell: Works for me. Thanks for clarifying.

deployments/cardano-nodes-env-staging.nix
deepfire: @jmitchell, the `-production` has no comparable code -- this was what I inherited.

deepfire:
..and the CI fails, because -- OOM.

deployments/infrastructure-env-production.nix
jmitchell: Ah, yes. They're separate deployments. 👍

deployments/cardano-nodes-env-staging.nix
domenkozar: Do we need `env:${environment}` if we already filter by deployment name?

modules/cardano.nix
domenkozar: String concatenation should be converted to anitquoting here `${if ... then .. else}` to prevent any errors in identation.

scripts/travis.sh
domenkozar: This alone won't be enough, best to skip CI for now and revert this change.

deployments/cardano-nodes-env-staging.nix
deepfire: @domenkozar, we want flexibility in subsetting.  Also, I think in the past, one of you guys argued for keeping `env:${environment}`.

domenkozar: I'm all for having both tags, but I don't think it makes sense to filter this query on both. You should either have a monitor per deployment (that would be nice, to be able to mute it) or per env, but not filter for both.

jmitchell: @domenkozar, some monitors should be used across all "production" environments, whether they're mainnet, testnet, infra, etc. A good example is the disk utilization monitor. We could just duplicate those monitors across all those envs, but it seems less error prone to use two tag types: one for deployment context (e.g. mainnet, testnet, staging, infra) and one to indicate severity (e.g. production/customer-impacting and internal). This would also enable monitors to distinguish the mainnet deployer from the staging deployer, both of which are infra, and set the escalation level accordingly.

domenkozar: We are still talking about two different things. What you and Serge refer to is the following line:

`services.dd-agent.tags = ["env:production", "depl:${config.deployment.name}"];`

That's good, actually we talked about setting this weeks ago.

The part that I'm confused is the above 15 line where we configure the monitor to filter on BOTH tags. My question is why both for this monitor that belong only to staging cardano nodes? (also `environment` shouldn't be used as a variable, but `staging` should be hardcoded as it makes no sense to have a var.

deepfire: @domenkozar, I think we want @jmitchell to tell us how he wants the monitors to be set up.. I don't have an argument either way.

domenkozar: Understood, I don't think this will break anythinig, it's just about being explicit what we want. Willing to merge if you can address the rest of the changes

domenkozar:
Also needs a rebase on top of staging due to recent staging PR merge.

scripts/travis.sh
deepfire: Done.

modules/cardano.nix
deepfire: Done.

deployments/cardano-nodes-env-staging.nix
domenkozar: But I'd still like `${environment}` -> `staging` change

deepfire: @domenkozar, the (unstated) reason why I was keeping it -- don't we want to ship that piece of code into some common part, eventually?

If you say that we don't want modularity at this level, I'm perfectly fine inlining the environment-sensitivity there.

deepfire: Personally, it just hurts seeing monitor definitions scattered around the code base.

deepfire:
All done, @domenkozar.

deepfire:
@domenkozar, this PR breaks evaluation in an unexpected way:
```
-- /nix/store/mb4rfbkliwjvcdcnqzgg4bgi45i8pb3s-nixops-1.6pre0_abcdef/bin/nixops deploy --show-trace --deployment rc-staging --max-concurrent-copy 50 -j 4 --include c-a-2
error: while evaluating anonymous function at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/options.nix:117:41, called from /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:199:52:                                                                                  while evaluating ‘scrubOptionValue’ at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/options.nix:113:22, called from /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/options.nix:117:44:                                                                                          while evaluating ‘isDerivation’ at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:295:18, called from /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/options.nix:114:8:                                                                                              while evaluating the attribute ‘elasticIPv4’ at /run/current-system/nixpkgs/lib/attrsets.nix:199:44:
while evaluating anonymous function at /run/current-system/nixpkgs/lib/modules.nix:75:45, called from /run/current-system/nixpkgs/lib/attrsets.nix:199:52:
while evaluating the attribute ‘value’ at /run/current-system/nixpkgs/lib/modules.nix:300:9:
while evaluating the option `deployment.ec2.elasticIPv4':
while evaluating the attribute ‘isDefined’ at /run/current-system/nixpkgs/lib/modules.nix:332:5:
while evaluating ‘filterOverrides’ at /run/current-system/nixpkgs/lib/modules.nix:407:21, called from /run/current-system/nixpkgs/lib/modules.nix:316:18:
while evaluating ‘concatMap’ at /run/current-system/nixpkgs/lib/lists.nix:79:18, called from /run/current-system/nixpkgs/lib/modules.nix:413:8:
while evaluating ‘concatMap’ at /run/current-system/nixpkgs/lib/lists.nix:79:18, called from /run/current-system/nixpkgs/lib/modules.nix:311:17:
while evaluating anonymous function at /run/current-system/nixpkgs/lib/modules.nix:311:28, called from undefined position:
while evaluating ‘dischargeProperties’ at /run/current-system/nixpkgs/lib/modules.nix:374:25, called from /run/current-system/nixpkgs/lib/modules.nix:312:62:
while evaluating the attribute ‘value’ at /run/current-system/nixpkgs/lib/modules.nix:204:48:
while evaluating the attribute ‘config.elasticIPv4’ at /home/staging/rc-staging/modules/network-wide.nix:7:7:
while evaluating the attribute ‘elasticIPs.c-a-1-ip’ at /nix/store/mb4rfbkliwjvcdcnqzgg4bgi45i8pb3s-nixops-1.6pre0_abcdef/share/nix/nixops/eval-machine-info.nix:97:3:
while evaluating ‘evalResources’ at /nix/store/mb4rfbkliwjvcdcnqzgg4bgi45i8pb3s-nixops-1.6pre0_abcdef/share/nix/nixops/eval-machine-info.nix:79:31, called from /nix/store/mb4rfbkliwjvcdcnqzgg4bgi45i8pb3s-nixops-1.6pre0_abcdef/share/nix/nixops/eval-machine-info.nix:97:26:                                                                                                               while evaluating ‘mapAttrs’ at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:198:17, called from /nix/store/mb4rfbkliwjvcdcnqzgg4bgi45i8pb3s-nixops-1.6pre0_abcdef/share/nix/nixops/eval-machine-info.nix:80:5:                                                                                                           while evaluating ‘zipAttrsWith’ at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:346:21, called from /nix/store/mb4rfbkliwjvcdcnqzgg4bgi45i8pb3s-nixops-1.6pre0_abcdef/share/nix/nixops/eval-machine-info.nix:97:58:                                                                                                      while evaluating ‘zipAttrsWithNames’ at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:331:33, called from /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:346:27:                                                                                       while evaluating ‘concatMap’ at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/lists.nix:79:18, called from /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:346:46:                                                                                                   while evaluating ‘zipAttrsWith’ at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:346:21, called from /nix/store/mb4rfbkliwjvcdcnqzgg4bgi45i8pb3s-nixops-1.6pre0_abcdef/share/nix/nixops/eval-machine-info.nix:69:21:                                                                                                      while evaluating ‘zipAttrsWithNames’ at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:331:33, called from /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:346:27:                                                                                       while evaluating ‘concatMap’ at /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/lists.nix:79:18, called from /nix/store/g30h5biv8cd443b5xsaax7ddc2giwpnw-nixpkgs-31a956f18c9251064997085fd7732562b1854726-src/lib/attrsets.nix:346:46:                                                                                                   value is a function while a set was expected, at undefined position
error: evaluation of the deployment specification failed
iohk-ops.hs: ProcFailed {procCommand = "/nix/store/mb4rfbkliwjvcdcnqzgg4bgi45i8pb3s-nixops-1.6pre0_abcdef/bin/nixops", procArguments = ["deploy","--show-trace","--deployment","rc-staging","--max-concurrent-copy","50","-j","4","--include","c-a-2"], procExitCode = ExitFailure 1}
```

PR #158, opened 2017-10-01
[DEVOPS-395] Document separation of concerns between files in deploym…

domenkozar:
…ents folder

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Collection of tools to deploy IOHK infrastructure.
 
 ### File structure
 
-- `deployments` - includes all NixOps deployments controlled via `.hs` scripts
+- `deployments` - includes all NixOps deployments inputs controlled via `.hs` scripts
 - `modules` - NixOS modules
 - `lib.nix` - wraps upstream `<nixpkgs/lib.nix>` with our common functions
 - `scripts` - has bash scripts not converted to Haskell/Turtle into Cardano.hs yet
@@ -25,11 +25,17 @@ resources that are only needed in all of deployments are provisioned.
 
 Per each set of machines (which can be combined or used standalone in a deployment) there are files:
 
-- `<{purpose}>.nix` set of "logical" machines
-- `<{purpose}-target-{target}.nix` set of physical machines for given target (aws, ...)
-- `<{purpose}-env-{environment}.nix` set of environment specific overrides (currently implies aws target)
+- `{purpose}.nix` set of "logical" machines
+- `{purpose}-target-{target}.nix` set of physical machines for given target (aws, ...)
+- `{purpose}-env-{environment}.nix` set of environment specific overrides (currently implies aws target)
 
-[Explanation about logical vs phystical can be found in Nixops manual](https://nixos.org/nixops/manual/#chap-introduction)
+[Explanation about logical vs physical can be found in Nixops manual](https://nixos.org/nixops/manual/#chap-introduction)
+
+Example for deploying cardano nodes in production:
+
+- cardano-nodes.nix
+- cardano-nodes-target-aws.nix
+- cardano-nodes-env-production.nix
 
 
 ### Getting SSH access
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 Collection of tools to deploy IOHK infrastructure.
 
-### Structure
+### File structure
 
 - `deployments` - includes all NixOps deployments controlled via `.hs` scripts
 - `modules` - NixOS modules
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,56 +1,36 @@
...
-     $ ./CardanoCSL.hs build
+    $(nix-build -A iohk-ops)/bin/iohk-ops --help
 
-### Deploying changes
 
-    $ ./CardanoCSL.hs deploy
+### Deployments file structure
 
-**Note this will not remove machines if they no longer exist.**
+There is one global resource file `deployments/keypairs.nix` for all deployments. That's where
```
jmitchell: That file doesn't exist on the target branch, but it does exist in `master`.

README.md
```diff
@@ -1,56 +1,36 @@
...
+resources that are only needed in all of deployments are provisioned.
 
-### Removing machines
+Per each set of machines (which can be combined or used standalone in a deployment) there are files:
 
-If you want to remove a machine simply remove them in the bottom part of the `nixops.nix` file.
-Then run `nixops deploy` with `-k`:
+- `<{purpose}>.nix` set of "logical" machines
+- `<{purpose}-target-{target}.nix` set of physical machines for given target (aws, ...)
+- `<{purpose}-env-{environment}.nix` set of environment specific overrides (currently implies aws target)
```
jmitchell: Can you use either `<` or `{` for meta-variables instead of both? The closing `>` is missing in the last two.

Can you also document examples of each meta-variable? For example:

* purpose: `cardano-nodes`
* target: `aws`
* environment: `production` or `staging`

README.md
```diff
@@ -1,56 +1,36 @@
...
 
-./Cardano.hs deploy
+[Explanation about logical vs phystical can be found in Nixops manual](https://nixos.org/nixops/manual/#chap-introduction)
```
jmitchell: s/phystical/physical/

README.md
domenkozar: Indeed, it has to be moved back.

domenkozar:
@jmitchell pushed a fix to address your concerns. About the hardcoding part, I'm not sure how to really phrase that. If there are files dedicated to environment overrides, why would it make sense to have variables scattered around the codebase?

deepfire:
As far as I'm concerned, this documents the effect of a design decision, but not its motivation.

domenkozar:
@deepfire but do you understand what are the implications of such design?

The motivations are the following:

a) have developers and devops managing clusters and thus the deployment interface should be stable and simple to understand, while it's internal implementation can evolve through time.

b) it should be easily verifiable what is going to be deployed (comparing environments and backends)

c) separation of concerns also need to imply how nixos module system is structured to be able to know what code goes into what file

What I still need to document is what goes into `modules/` folder.

README.md
jmitchell: That should be done before this is merged.

deepfire:
@domenkozar, first about the motivations -- I agree with those motivations, but from your description it's not very clear how they help us choose one of two options over the other.  So let me try to have a first stab at clarification.

Also, please note that we have two different large questions -- physical spec separation and environment specificity -- of which I am only going to address the latter -- for now.

First, let me explicitly state the _two options that we have_, with regard to _environment-specificity_ -- as I personally understand them:

Option A.  
Have generalized declarations that:
    - are defined in terms of a set of network-wide options -- that serve as the external configuration interface 
    - are strongly grouped by implementation aspect (like security groups, datadog monitors, for example)

Option B. 
Have environment-specialized declarations that:
    - provide environment-concrete definitions for the implementation aspects (like security groups, datadog monitors)
    - effectively have these implementation aspects scattered across:
        - the set of env-specific parts
        - the generalized part

Having spelled the options out, let me go over your list and see how it swings with their regard:

> a) have developers and devops managing clusters and thus the deployment interface should be stable and simple to understand, while it's internal implementation can evolve through time.

Option A: The network-wide options set of options serves as the deployment interface that is, perhaps, simplest to understand.  At least as long as the options are chosen wisely and are properly documented.  In case documentation isn't sufficient, the implementation aspect (SG, DD, etc) can be inspected comfortably, since its implementation resides in a single place.

Option B: There is no abstract option exposed, so the entire implementation has to be used as the deployment interface -- and it's scattered between the generalized (`modules/*.nix`), and env-specific parts.

My call here is that Option B loses.

> b) it should be easily verifiable what is going to be deployed (comparing environments and backends)

It's not really clear what is meant by "verifiable".  `nixops deploy --debug` prints out an entire machine spec that is to be deployed.

I am ignoring the back-end aspect for now (as I said previously), and so I'm going to interpret that as "comparing environments".  I am also going to interpret that even more narrowly, and assume that we want to interpret a single aspect between environments.

Option A:
- shallow comparison between environments amounts to simply comparing the two sets of network-wide _settings_ relevant to a particular aspect -- based on the names and documentation of those settings.
- "deep" comparison between environments amounts to _additionally_ opening a single _aspect-specific_ implementation file and inspecting how these network-wide options affect the expressions.

Option B:
- shallow comparison is impossible: you'd need to inspect the full set of files involved: the generalized part and the two environment specific _implementations_
- "deep" comparison between environments amounts to analyzing three implementation files. 

My call here is that Option B loses.

> c) separation of concerns also need to imply how nixos module system is structured to be able to know what code goes into what file

Domen, I'm afraid I don't quite understand what you refer here to -- and so I'm even less equipped to see how it affects the two options I outlined.

Finally for my part, I would say that the list you provided is missing a point that I find crucial:

4. Help understanding the system as a whole.

Understanding the system as a whole is tremendously important for its long-term health. 

Here, the fact that implementation aspects are scattered in the option B doesn't help it at all.

PR #159, opened 2017-10-03
[DEVOPS-227] pass `+RTS -p -RTS` when profiling is enabled

cleverca22:


modules/cardano-service.nix
```diff
@@ -6,7 +6,7 @@ let
   cfg = config.services.cardano-node;
   name = "cardano-node";
   stateDir = "/var/lib/cardano-node";
-  cardano = (import ./../default.nix {}).cardano-sl-static;
+  cardano = (import ./../default.nix { enableProfiling = config.services.cardano-node.enableProfiling; }).cardano-sl-static;
```
domenkozar: You can use `cfg`

PR #162, opened 2017-10-06
Devops 377: add `reallocate-core-ips` subcommand to `iohk-ops`

deepfire:


iohk/NixOps.hs
```diff
@@ -408,7 +408,7 @@ readTopology file = do
     Left err -> errorT $ format ("Failed to parse topology file: "%fp%": "%w) file err
 
 data SimpleTopo
-  =  SimpleTopo (Map.Map NodeName SimpleNode)
+  =  SimpleTopo { fromSimpleTopo :: (Map.Map NodeName SimpleNode) }
```
domenkozar: Shouldn't this be newtype?

iohk/NixOps.hs
```diff
@@ -445,8 +445,11 @@ instance ToJSON NodeType
 topoNodes :: SimpleTopo -> [NodeName]
 topoNodes (SimpleTopo cmap) = Map.keys cmap
 
+topoCores :: SimpleTopo -> [NodeName]
+topoCores = (fst <$>) . filter ((== NodeCore) . snType . snd) . Map.toList . fromSimpleTopo
+
 topoNCores :: SimpleTopo -> Int
-topoNCores (SimpleTopo cmap) = Map.size $ flip Map.filter cmap ((== NodeCore) . snType)
+topoNCores = length . topoCores
```
domenkozar: This function seems redundant now

iohk/iohk-ops.hs
```diff
@@ -181,6 +182,8 @@ centralCommandParser =
    , ("destroy",                "Destroy the whole cluster",                                        pure Destroy)
    , ("delete",                 "Unregistr the cluster from NixOps",                                pure Delete)
    , ("fromscratch",            "Destroy, Delete, Create, Deploy",                                  pure FromScratch)
+   , ("reallocate-core-ips",    "Destroy elastic IPs corresponding to the nodes listed and reprovision cluster",
```
domenkozar: Reprovision means that resources would be reprovisioned, we should just say "deploy"

iohk/NixOps.hs
deepfire: It started as a `newtype`, right -- but for some reason I had to turn this into a `data`.  Don't remember why : -/

Besides, there is no _real_ difference (except performance, which I think we can agree is not an issue here) between these options.

iohk/NixOps.hs
deepfire: You seem to refer to the fact that `topoNCores` should now be inlined (in two locations, as of now)?

I find it arguable, but OK.  Done.

iohk/iohk-ops.hs
deepfire: Done.

domenkozar:
Lastly, this needs a section in operational manual. What is the symptom and what to run as a solution (this new command).

deepfire:
@domenkozar, what do you refer to by symptoms -- does that mean assessment of whether or not we are under a DDoS?  That would probably include DataDog metrics (some of which we might already have?).

If this is what you mean, do we already have a DDoS response section in the manual?

Again, if this is what you mean, I think we might need to extend the scope of https://issues.serokell.io/issue/DEVOPS-377 to cover documentation of DDoS response -- right now it's clearly not in the task description.

domenkozar:
Ah, this wasn't made against staging. I'll reopen.

PR #164, opened 2017-10-08
[DEVOPS-413] development clusters

domenkozar:


domenkozar:
Changes for development cluster seem good, but we should leave out cardano-sl bump and configuration.yaml unless we properly merge those upstream.

topology-development.yaml
```diff
@@ -3,42 +3,52 @@ nodes:
...
     type: relay
     region: ap-southeast-1
     zone:   ap-southeast-1b
-    static-routes: [["a1"]]
-    addr: r1.cardano
+    static-routes: [["a1"], ["b1"]]
+    host: r1.cardano
     port: 3000
+    kademlia: false
   'r2':
```
domenkozar: We only need one relay really

cardano-sl-src.json
```diff
@@ -1,6 +1,6 @@
 {
     "url": "https://github.com/input-output-hk/cardano-sl",
     "fetchSubmodules": "true",
-    "sha256": "0hrfha4xsvdxhfnkpjc3bclijsixwibnckvvb9mbf8kc563dv61f",
-    "rev": "f2e1da159543b439f30ecef2d288d4661d6765a5"
+    "sha256": "0a480p4wcjj8hi7bcgkv85jy94ad59isfg7lfnlk8qsv0q2bj1wh",
+    "rev": "7f5aff98f39ccf50ec3f5c98638e93535f16021c"
```
domenkozar: should be removed

configuration.yaml.tmpl
```diff
@@ -14955,3 +14955,26 @@ testnet_staging_wallet_macos64: &testnet_staging_wallet_macos64
     <<: *mainnet_base_update
     applicationName: csl-daedalus
     systemTag: macos64
+
+##############################################################################
```
domenkozar: to be removed

domenkozar:
I'll make the necessary changes here

deepfire:
@domenkozar, I think we also need a configuration.yaml update, let me check..

domenkozar:
@deepfire we are going to use `devnet` for development clusters when it's all ready, I've removed the configuration.yaml changes that were specific to the branch.

deepfire:
@domenkozar, I think @gromakovsky already has a `devnet` configuration key prepared somewhere.. we might include it here as well -- and update `iohk-ops` to use it.

gromakovsky:
Right, `devnet` is in master.

PR #170, opened 2017-10-10
[DEVOPS-421] rc-Staging deploy of cardano-sl-1.0 HEAD

cleverca22:


nixpkgs-src.json
```diff
@@ -1,6 +1,6 @@
 {
-    "owner":  "NixOS",
-    "repo":   "nixpkgs",
-    "rev":    "31a956f18c9251064997085fd7732562b1854726",
-    "sha256": "19xpdpbw7yn9im6npdr3pdjl2bgkj07k5w0qwymlpxksk4p77nah"
-}
+    "url": "https://github.com/nixos/nixpkgs",
```
jmitchell: Is replacing `owner` and `repo` with `url` intentional? I'm not picky if either works; just checking.

cleverca22: it failed travis because of that change, i believe its a bug in `iohk-ops set-rev`

domenkozar: cc @deepfire

PR #171, opened 2017-10-11
Devops 410 iohk ops

deepfire:
`nix-shell` will alias `io` to a binary build of `iohk-ops`
`nix-shell --arg io-dev-mode true` will alias `io` to an interpreted version

All variants source `iohk-ops` from the current definition.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -40,6 +40,7 @@ in {
              (compiler.callPackage ./iohk/default.nix {})
              (drv: {
                 executableToolDepends = [ pkgs.makeWrapper ];
+                libraryHaskellDepends = iohk-ops-extra-runtime-deps;
                 postInstall = ''
                   wrapProgram $out/bin/iohk-ops \
                   --prefix PATH : "${pkgs.lib.makeBinPath iohk-ops-extra-runtime-deps}"
--- a/default.nix
+++ b/default.nix
@@ -15,6 +15,9 @@ with pkgs.haskell.lib;
 let
   iohk-ops-extra-runtime-deps = [
     pkgs.git pkgs.nix-prefetch-scripts compiler.yaml
+    pkgs.wget pkgs.awscli # for scripts/aws.hs
+    pkgs.nodejs           # for cardano-sl/scripts/js/genesis-hash.js
+                          #     ..which also needs `npm install blakejs canonical-json`
   ];
   # we allow on purpose for cardano-sl to have it's own nixpkgs to avoid rebuilds
   cardano-sl-src = builtins.fromJSON (builtins.readFile ./cardano-sl-src.json);
--- a/shell.nix
+++ b/shell.nix
@@ -1,5 +1,6 @@
 { localLib ? import ./lib.nix
 , pkgs ? import (localLib.fetchNixPkgs) {}
+, io-dev-mode ? false
 }:
 
 let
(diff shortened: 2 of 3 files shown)


shell.nix
```diff
@@ -1,12 +1,23 @@
...
     ];
-  drv = pkgs.haskell.lib.overrideCabal
-         (import ./default.nix {}).iohk-ops
-         (drv: { libraryHaskellDepends = extraDeps; });
-in drv.env
+  iohk-ops = pkgs.haskell.lib.overrideCabal
+             (import ./default.nix {}).iohk-ops
+             (drv: { libraryHaskellDepends = extraDeps; });
+  ioAlias = if io-dev-mode
+            then ''runhaskell -iiohk iohk/iohk-ops.hs''
```
domenkozar: So we can now remove `iohk/iohk-env.nix`?

shell.nix
```diff
@@ -1,12 +1,23 @@
...
-    [ wget awscli nodejs # for scripts/aws.hs
+    [ wget awscli  # for scripts/aws.hs
+      nodejs       # for cardano-sl/scripts/js/genesis-hash.js
+                   #     ..which also needs `npm install blakejs canonical-json`
     ];
-  drv = pkgs.haskell.lib.overrideCabal
-         (import ./default.nix {}).iohk-ops
-         (drv: { libraryHaskellDepends = extraDeps; });
-in drv.env
+  iohk-ops = pkgs.haskell.lib.overrideCabal
```
domenkozar: This really belongs to `default.nix`

deepfire: Yes, agreed..

domenkozar:
Also first time in AC is not fullfilled since cardano-deployer doesn't have iohk-ops installed globally

shell.nix
deepfire: Tough question -- I hoped we could retain the ability to use `io` outside `nix-shell`.

But if you deem it necessary, sure.

deepfire:
Wow, I thought it is. Hmm.

deepfire:
Apparently it's a regression.  Hydra master and slaves do have it installed, and I don't remember removing it from the deployer.. Heh.

default.nix
```diff
@@ -15,6 +15,9 @@ with pkgs.haskell.lib;
 let
   iohk-ops-extra-runtime-deps = [
     pkgs.git pkgs.nix-prefetch-scripts compiler.yaml
+    pkgs.wget pkgs.awscli # for scripts/aws.hs
+    pkgs.nodejs           # for cardano-sl/scripts/js/genesis-hash.js
```
domenkozar: This does now work with `io` alias? I'd assume we need to add it to a dependency list `libraryDepends` or similar

deepfire: Good question, I was contemplative and unsure when making this change!

Now that I think about it more -- this should work for the non-interactive version, because that's got its `wrapProgram` scripts generated.

It would, however fail for the `runhaskell` version -- only in case with `pkgs.nodejs`, when we're regenerating genesis.

deepfire: I guess I'll add all of those to `libraryHaskellDepends`, as it was implemented previously.

deepfire: Done.

domenkozar:
This looks good, but I think aliases don't work with nix-shell:

```
$ nix-shell --run "io --help"
/run/user/499/nix-shell.w9NpCd/rc: line 1: io: command not found
```

I have yet to figure out why.

domenkozar:
Filed https://github.com/NixOS/nix/issues/1598, I'll push a fix with a bash function that does work.

deepfire:
Heh, "rebase and merge" is a merge with fast forwarding.  Did not expect that : -)

PR #172, opened 2017-10-12
[DEVOPS-77] create a tool that can map sshd logs to a user

cleverca22:
if this tool is ran with `$keys` set to the right path, and is in the group `systemd-journald`, it will monitor sshd logs, and report messages like the following
```
ssh-audit: ALERT, unknown ssh keypair
ssh-audit: attribute: michaelKey user: michael.bishop
```

it will report those immediately after `sshd: Accepted publickey for user from a.b.c.d port x ssh2: RSA:...`

papertrail can then be configured to record both of them together

i'll add the nixos module config tomorrow


ssh-audit/ssh-audit.hs
```diff
@@ -0,0 +1,108 @@
...
+hashPubKey :: ByteString -> Either String ByteString
+hashPubKey pubkey = Base64.encode . hashSHA256 <$> Base64.decode pubkey
+
+parseLine :: ByteString -> Either String PubKey
+parseLine line = do
+  (attrName, keyType, pubkey, rawUser) <- do
+    case BSC.words line of
+      (a:k:p:ru) -> pure (a, k, p, ru)
+      other      -> Left $ "unable to parse " <> show other
+  let user = BSC.unwords rawUser
```
vizanto: will this work with user "Alex Vieth" ? (note the space)

PR #174, opened 2017-10-13
[DEVOPS-384] Use genesis/configuration.yaml from cardano-sl repo

domenkozar:
Path should be correct:

```
$ nix-repl default.nix 
Welcome to Nix version 1.11.15. Type :? for help.

Loading ‘default.nix’...
Added 371 variables.

nix-repl> cardano-sl.src
/nix/store/x2isfbhprq0z3lx3anj495d7qgqgvi7w-cardano-sl-824b230/node

$ ls /nix/store/x2isfbhprq0z3lx3anj495d7qgqgvi7w-cardano-sl-824b230/node
bench               LICENSE                                        mainnet-genesis.json  test
cardano-sl.cabal    mainnet-genesis-dryrun-with-stakeholders.json  README.md
configuration.yaml  mainnet-genesis-internal-dryrun.json           src
```


default.nix
```diff
@@ -16,8 +16,6 @@ let
   iohk-ops-extra-runtime-deps = [
     pkgs.git pkgs.nix-prefetch-scripts compiler.yaml
     pkgs.wget pkgs.awscli # for scripts/aws.hs
-    pkgs.nodejs           # for cardano-sl/scripts/js/genesis-hash.js
```
deepfire: What is the new way to perform the manual routine, without `npm` being available?

deepfire:
@domenkozar, I think we need to at least document the manual routine, if we're removing automation.

default.nix
domenkozar: I don't see we need to run this as part of iohk-ops anymore, but we could nixify the upstream

PR #177, opened 2017-10-16
[DEVOPS-453] Development topology fixes

deepfire:


globals.nix
```diff
@@ -74,7 +74,7 @@ let topologySpec     = builtins.fromJSON (builtins.readFile topologyFile);
     relays          = filter     (x: x.value.typeIsRelay)             indexed;
     nodeMap         = listToAttrs (cores ++ relays);
     # WARNING: this depends on the sort order, as explained above.
-    firstRelay      = findFirst (x: x.value.typeIsRelay) (throw "No relays in topology, it's sad we can't live, but then.. who does?") indexed;
+    firstRelay      = findFirst (x: x.value.typeIsRelay) ({ name = "fake-non-relay"; value = { type = "relay"; i = builtins.length indexed; }; }) indexed;
```
domenkozar: why is suddently there a need for fake relay where previously it was an evaluation error?

default.nix
```diff
@@ -38,10 +38,10 @@ in {
              (compiler.callPackage ./iohk/default.nix {})
              (drv: {
                 executableToolDepends = [ pkgs.makeWrapper ];
-                libraryHaskellDepends = iohk-ops-extra-runtime-deps;
+                libraryHaskellDepends = iohk-ops-extra-runtime-deps ++ [ cardano-sl-pkgs.cardano-sl-auxx pkgs.file nixops ];
```
domenkozar: Why are these needed to be in nix-shell?

deepfire: auxx and file is for the update proposals
keygen is for dev cluster deployments

globals.nix
deepfire: we can evaluate without nodes now

domenkozar: But shouldn't we skip the whole topology evaluation if there are no nodes?

deepfire: Ideally, yes.

deepfire:
All issue ids have been added.

default.nix
```diff
@@ -38,10 +41,10 @@ in {
              (compiler.callPackage ./iohk/default.nix {})
              (drv: {
                 executableToolDepends = [ pkgs.makeWrapper ];
-                libraryHaskellDepends = iohk-ops-extra-runtime-deps;
+                libraryHaskellDepends = iohk-ops-extra-runtime-deps ++ [ nixops ];
```
domenkozar: shouldn't nixops be part of `iohk-ops-extra-runtime-deps` as well?

deepfire: Done..

deepfire:
@domenkozar, the nodeless-ness PR was separated into https://github.com/input-output-hk/iohk-ops/pull/182

default.nix
```diff
@@ -34,6 +23,23 @@ in {
         sha256 = "1fly6ry7ksj7v5rl27jg5mnxdbjwn40kk47gplyvslpvijk65m4q";
       };
     in (import "${nixopsUnstable}/release.nix" {}).build.${system};
+  iohk-ops-extra-runtime-deps = [
+    pkgs.gitFull pkgs.nix-prefetch-scripts compiler.yaml
```
domenkozar: Do we really need gitFull?

domenkozar: It only adds 50MB, so looks OKish.

shell.nix
```diff
@@ -11,7 +11,7 @@ let
 in pkgs.lib.overrideDerivation iohk-ops.env
    (old: {
      shellHook = ''
-       io () {
+       function io {
```
domenkozar: @deepfire what's the reason for this change? If I understand correctly `io ()` is POSIX.

cleverca22: i also ran into this failing on the deployer, couldnt figure out why it was a problem

deepfire:
I can add an arbitrary commit id -- it's really a generic cosmetic thing, not attached anywhere and not meriting an issue of its own.

shell.nix
domenkozar: Can you paste the error?

domenkozar: Aha was it:

```
$ nix-shell
bash: eval: line 69: syntax error near unexpected token `('
bash: eval: line 69: `io () {'
``

cleverca22: yep, exactly

domenkozar: Really strange, bash should support this syntax.

PR #180, opened 2017-10-20
 [DEVOPS-455] implement S3 bucket storage for report-server logs

cleverca22:
currently, its enabled for both mainnet and staging, but it can easily be adjusted to target just one

due to boto2 in nixops, the bucket must be in a region that supports auth method version 2
the report-server ec2 is currently in frankfurt, so latency to the bucket may be high

### DIFF
--- a/deployments/report-server-bucket-storage.nix
+++ b/deployments/report-server-bucket-storage.nix
@@ -4,6 +4,7 @@
   # this file will create an S3 bucket, and IAM role with access to that bucket
   # then give the report-server machine access to that role, and mount the bucket within the instance
   report-server = { pkgs, resources, ... }: {
+    # nixops can only set this at creation, it will need to be manualy set
     deployment.ec2.instanceProfile = resources.iamRoles.report-role.name;
     fileSystems."/var/lib/report-server" = {
       # mount unsets PATH when running this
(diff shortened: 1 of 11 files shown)


deployments/report-server-bucket-storage.nix
```diff
@@ -0,0 +1,60 @@
...
+      device = "/run/current-system/sw/bin/s3fs#${resources.s3Buckets.report-bucket-3.name}";
+      fsType = "fuse";
+      options = [ "_netdev" "allow_other" "iam_role" ];
+    };
+    environment.systemPackages = [ pkgs.s3fs ];
+  };
+  resources = {
+    s3Buckets = {
+      # if moved to another region, you must rename the bucket
+      report-bucket-3 = { config, uuid, ... }: {
```
domenkozar: Hmm, we should probably pick a better name :) Maybe `report-server-logs`?

cleverca22: also keep in mind that nixops will mangle the name a bit, https://github.com/NixOS/nixops/blob/master/nix/s3-bucket.nix#L10

domenkozar:
Blocked on DEVOPS-469 to be able to migrate using frankfurt region.

PR #185, opened 2017-11-01
release/mainnet 2017-11-01

deepfire:
The branch is not pretty, and it deserved dis-entangling before PR creation.

..but we also want to record what was deployed.

### DIFF
--- a/deployments/report-server-bucket-storage.nix
+++ b/deployments/report-server-bucket-storage.nix
@@ -10,7 +10,7 @@
       # mount unsets PATH when running this
       device = "/run/current-system/sw/bin/s3fs#${resources.s3Buckets.report-server-logs.name}";
       fsType = "fuse";
-      options = [ "_netdev" "allow_other" "iam_role" ];
+      options = [ "_netdev" "allow_other" ("iam_role=${resources.iamRoles.report-role.name}") ];
     };
     environment.systemPackages = [ pkgs.s3fs ];
   };
--- a/cardano-sl-src.json
+++ b/cardano-sl-src.json
@@ -1,6 +1,6 @@
 {
     "url": "https://github.com/input-output-hk/cardano-sl",
     "fetchSubmodules": "true",
-    "sha256": "0glgrw6dgxxflxhk1wmylrj2vrgk9apqd4q3xyf1dvc51j1rhg5n",
-    "rev": "824b2303a5dd0058e617a351a1667fa66973d919"
+    "sha256": "0gpsv7qji70k54wm1hw8sak1sq2cfdaynh6s027r6rksggin0vv7",
+    "rev": "8b127d61e0a70f2dcb5287957871b888459421d6"
 }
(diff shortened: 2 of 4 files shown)


deployments/cardano-nodes-env-production.nix
```diff
@@ -15,7 +15,7 @@ let nodeMap = globals.nodeMap; in
       ntp = mkMonitor ntp_monitor;
       cardano_node_simple_process = mkMonitor cardano_node_simple_process_monitor;
       chain_quality = mkMonitor chain_quality_monitor;
-      failed_cherish_loop = mkMonitor failed_cherish_loop_monitor;
+      # failed_cherish_loop = mkMonitor failed_cherish_loop_monitor;
```
domenkozar: This should be uncommented now

deployments/cardano-nodes-env-staging.nix
```diff
@@ -39,7 +39,7 @@ let nodeMap = globals.nodeMap; in
       });
 
       chain_quality = mkMonitor chain_quality_monitor;
-      failed_cherish_loop = mkMonitor failed_cherish_loop_monitor;
+      # failed_cherish_loop = mkMonitor failed_cherish_loop_monitor;
```
domenkozar: same here

staging-testnet.yaml
```diff
@@ -19,6 +19,7 @@ gen-cmdline: template -c staging-testnet.yaml --environment staging --nixops /ni
 environment: Staging
 topology: topology-staging.yaml
 name: rc-staging
+installer-bucket: update-system-testing
```
domenkozar: This doesn't belong here, but is part of #183

deployments/report-server-bucket-storage.nix
```diff
@@ -10,7 +10,7 @@
       # mount unsets PATH when running this
       device = "/run/current-system/sw/bin/s3fs#${resources.s3Buckets.report-server-logs.name}";
       fsType = "fuse";
-      options = [ "_netdev" "allow_other" "iam_role" ];
+      options = [ "_netdev" "allow_other" "nonempty" ("iam_role=${resources.iamRoles.report-role.name}") ];
```
domenkozar: Do we really need nonemtpy? I would imagine the folder should be empty at the mount time. Probably report-server itself should depend on this mount to finish first. 

cc @cleverca22

deepfire: @domenkozar, it's more reliable, I would say -- should anything land there for any reason, the mount will still succeed.

But I'm also agnostic -- whatever you like.

domenkozar:
If there's anything in that folder, it's misconfiguration. We should make sure that report server starts after S3 has been mounted. Then S3 mounting should make sure the folder is empty, otherwise we're missing reports.

deepfire:
@domenkozar, dropped "nonempty".

domenkozar:
https://github.com/input-output-hk/iohk-ops/blob/develop/modules/report-server.nix#L69 already exists, so looks good!

@deepfire last thing is resolving git conflicts :)

deepfire:
@domenkozar, now supplied from conflict-free areas only!

PR #186, opened 2017-11-01
[DEVOPS-482] Sync nixpkgs pinning

domenkozar:


fetch-nixpkgs.nix
```diff
@@ -1,32 +1,23 @@
...
-    inherit input;
-    coreutils = builtins.storePath nixcfg.coreutils;
-    tar = builtins.storePath nixcfg.tar;
-    gzip = builtins.storePath nixcfg.gzip;
-  };
-  fetchNixpkgsTarWithNix = let
-    spec = builtins.fromJSON (builtins.readFile ./nixpkgs-src.json);
-  in import <nix/fetchurl.nix> {
+  spec = builtins.fromJSON (builtins.readFile ./nixpkgs-src.json);
+  src = import <nix/fetchurl.nix> {
```
deepfire: Interesting, I couldn't find where `<nix/fetchurl.nix>` comes from..

fetch-nixpkgs.nix
```diff
@@ -1,32 +1,23 @@
...
     url = "https://github.com/${spec.owner}/${spec.repo}/archive/${spec.rev}.tar.gz";
     inherit (spec) sha256;
   };
-in
-  unpackTar fetchNixpkgsTarWithNix
+  nixcfg = import <nix/config.nix>;
```
deepfire: Ditto..

deepfire:
So this is a minimal, bare-bones derivation that fetches and unpacks Nixpkgs without depending on Nixpkgs per se -- but instead getting its dependencies (`gzip`, `coreutils`, etc.) from Nix's configuration itself, somehow.

Looks very nice!

fetch-nixpkgs.nix
domenkozar: It's a search path for Nix installation: https://github.com/NixOS/nix/tree/master/corepkgs

PR #187, opened 2017-11-02
Devops 157 support developers

deepfire:
- dev-env key generation
- DNS provisioning in the `aws.iohkdev.io` domain

### DIFF
--- a/iohk/NixOps.hs
+++ b/iohk/NixOps.hs
@@ -84,7 +84,7 @@ module NixOps (
   , URL(..)
   , Username(..)
   , Zone(..)
-  
+
   -- * Flags
   , BuildOnly(..)
   , DoCommit(..)
@@ -97,7 +97,7 @@ module NixOps (
   , flag
   , toBool
 
-  -- 
+  --
   , parserBranch
   , parserCommit
   , parserNodeLimit
@@ -576,6 +576,7 @@ inprocs bin args inp = do
 
 cmd   :: Options -> Text -> [Text] -> IO ()
 cmd'  :: Options -> Text -> [Text] -> IO (ExitCode, Text)
+cmd'' :: Options -> Text           -> IO (ExitCode, Text, Text)
 incmd :: Options -> Text -> [Text] -> IO Text
 
 cmd   Options{..} bin args = do
@@ -584,6 +585,9 @@ cmd   Options{..} bin args = do
 cmd'  Options{..} bin args = do
   when (toBool oVerbose) $ logCmd bin args
   Turtle.procStrict bin args empty
+cmd'' Options{..} command  = do
+  when (toBool oVerbose) $ logCmd command []
+  Turtle.shellStrictWithErr command empty
 incmd Options{..} bin args = do
   when (toBool oVerbose) $ logCmd bin args
   inprocs bin args empty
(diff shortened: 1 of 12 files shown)


iohk/iohk-ops.hs
```diff
@@ -289,4 +289,44 @@ runTemplate o@Options{..} Template{..} args = do
...
+      sh $ do
+        k <- Turtle.find ((prefix "keys/keys-testnet/rich/key") <> (suffix ".sk"))
+          "keys/keys-testnet/rich"
+        cp k $ "keys" </> Path.filename k
+  echo "Cluster deployment has been prepared."
+
+runNew _ _ _ = error "impossible"
+
+-- | Use 'cardano-keygen' to create keys for a develoment cluster.
+generateDevKeys :: Options -> ConfigurationKey -> Turtle.FilePath -> IO ()
```
domenkozar: This should be named `generateGenesis` as a general case, then used for development clusters as needed.

modules/development.nix
```diff
@@ -7,8 +7,8 @@ with import ./../lib.nix;
     global = {
       allocateElasticIP            = false;
       enableEkgWeb                 = true;
-      dnsHostname                  = null;
-      dnsDomainname                = mkForce null;
+      dnsHostname                  = "${name}-${config.deployment.name}";
```
domenkozar: `name` has the same value as `config.deployment.name`, did you verify the resulting URL?

iohk/iohk-ops.hs
deepfire: The genesis is generated by `cardano-node` during startup, so it's really just generation of keys compatible with a provided configuration key -- to the best of my understanding.

cc @gromakovsky

modules/development.nix
deepfire: @domenkozar, are you sure?

I'm reading https://nixos.org/nixops/manual/#ssec-relnotes-1.3 differently.

Also: https://github.com/NixOS/nixops/blob/master/nix/eval-machine-info.nix#L73

deepfire: As to the DNS provisioning -- I've tested it, indeed.

domenkozar: You're right, `name` is machine name.

iohk/iohk-ops.hs
domenkozar: Why is there parameter called `--genesis-out-dir`?

In that case we should give it a bit descriptive name, like `generateStakeKeys`.

gromakovsky: Probably because @pva701 is bad at naming things :)
In master we also have `dump-genesis-data --path PATH` command.
I can rename `generate-keys-by-spec` to `dump-genesis-secrets` (not keys, because it also dumps AVVM seeds) and its path argument to `--path`. I'd like to do it in master, because it's not something strictly necessary for 1.0.

deepfire:
@domenkozar, CI is green now.

domenkozar:
@deepfire is it complete so I can re-review? Note that it has git conflicts.

deepfire:
@domenkozar, fixed the `cardano-sl-src.json` bumping conflict.

Answering your question -- I believe I have addressed all of your review comments, yes.

iohk/NixOps.hs
```diff
@@ -820,29 +519,33 @@ cmd   Options{..} bin args = do
...

 -- * Invoking nixops
 --
-iohkNixopsPath :: FilePath -> FilePath
-iohkNixopsPath defaultNix =
-  let storePath  = Sys.unsafePerformIO $ inprocs "nix-build" ["-A", "nixops", format fp defaultNix] $
+iohkNixopsPath :: FilePath
+iohkNixopsPath =
+  let defaultNix = "default.nix"
+      storePath  = Sys.unsafePerformIO $ inprocs "nix-build" ["-A", "nixops", format fp defaultNix] $
```
domenkozar: Why not just make it an `IO` function?

iohk/NixOps.hs
```diff
@@ -820,29 +519,33 @@ cmd   Options{..} bin args = do
...
-  executor o (format fp $ iohkNixopsPath "default.nix")
-  (fromCmd com : nixopsCmdOptions o c <> fmap fromArg args)
+  executor o (format fp iohkNixopsPath)
+    (fromCmd com : nixopsCmdOptions o c <> fmap fromArg args)
 
-nixops' :: Options -> NixopsConfig -> NixopsCmd -> [Arg] -> IO (ExitCode, Text)
-nixops  :: Options -> NixopsConfig -> NixopsCmd -> [Arg] -> IO ()
-nixops' = nixops'' cmd'
-nixops  = nixops'' cmd
+nixops'  :: Options -> NixopsConfig -> NixopsCmd -> [Arg] -> IO (ExitCode, Text)
```
domenkozar: It's really really hard to follow functions like this. There are 5 functions used in below two lines with almost indentical names. Why not give them appropriate names?

iohk/iohk-ops.hs
domenkozar: I'd still call this function `generateStakeKeys` as that's what it does, the fact that it's developer environment comes from it's usage, not definition.

iohk/NixOps.hs
deepfire: The intent here is to calculate it just once -- an IO action would be recomputed repeatedly.

iohk/NixOps.hs
deepfire: I believe I didn't change this code in this PR.  Can we address this later?

deepfire: Also, do you have any suggestions?

iohk/NixOps.hs
```diff
@@ -820,29 +519,33 @@ cmd   Options{..} bin args = do
...
-iohkNixopsPath defaultNix =
-  let storePath  = Sys.unsafePerformIO $ inprocs "nix-build" ["-A", "nixops", format fp defaultNix] $
+iohkNixopsPath :: FilePath
+iohkNixopsPath =
+  let defaultNix = "default.nix"
+      storePath  = Sys.unsafePerformIO $ inprocs "nix-build" ["-A", "nixops", format fp defaultNix] $
                    (trace (T.unpack $ format ("INFO: using "%fp%" expression for its definition of 'nixops'") defaultNix) empty)
-      nixopsPath = Path.fromText $ T.strip storePath <> "/bin/nixops"
-  in trace (T.unpack $ format ("INFO: nixops is "%fp) nixopsPath) nixopsPath
+      opsPath    = Path.fromText $ T.strip storePath <> "/bin/nixops"
```
domenkozar: We have `nixops` and `iohk-ops`, so `ops` is an overloaded term, why make the name binding ambigious?

iohk/NixOps.hs
domenkozar: Sure

iohk/NixOps.hs
deepfire: This is a local variable, and the meaning is very clear, IMO -- its usage is really spread across three lines.   But ok.

PR #191, opened 2017-11-15
[DEVOPS-382] staging infra

deepfire:
- slight massage of the infrastructure definitions
- addition of the staging infra

### DIFF
--- a/deployments/infrastructure-env-staging.nix
+++ b/deployments/infrastructure-env-staging.nix
@@ -13,6 +13,10 @@ in {
       ./../modules/datadog.nix
     ];
 
+    ## 
+    security.acme.certs = mkForce {};
+    services.nginx.enable = mkForce false;
+
     services.dd-agent.tags = ["env:staging" "depl:${config.deployment.name}"];
 
     deployment.ec2.elasticIPv4 = resources.elasticIPs.hydra-ip;
--- a/deployments/infrastructure-env-staging.nix
+++ b/deployments/infrastructure-env-staging.nix
@@ -1,0 +1,63 @@
+{ IOHKaccessKeyId, ... }:
+
+with (import ./../lib.nix);
+let region = "eu-central-1";
+    accessKeyId = IOHKaccessKeyId;
+in {
+  network.description = "IOHK infrastructure staging";
+
+  hydra = { config, pkgs, resources, ... }: {
+
+    imports = [
+      ./../modules/papertrail.nix
+      ./../modules/datadog.nix
+    ];
+
+    services.dd-agent.tags = ["env:staging" "depl:${config.deployment.name}"];
+
+    deployment.ec2.elasticIPv4 = resources.elasticIPs.hydra-ip;
+
+    deployment.route53.accessKeyId = config.deployment.ec2.accessKeyId;
+    deployment.route53.hostName = "hydra.awstest2.iohkdev.io";
+  };
+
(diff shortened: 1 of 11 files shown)


iohk/NixOps.hs
```diff
@@ -585,8 +574,14 @@ computeFinalDeploymentArgs o@Options{..} NixopsConfig{..} = do
 
 modify :: Options -> NixopsConfig -> IO ()
 modify o@Options{..} c@NixopsConfig{..} = do
-  printf ("Syncing Nix->state for deployment "%s%"\n") $ fromNixopsDepl cName
-  nixops o c "modify" $ Arg <$> cFiles
+  deplExists <- exists o c
```
domenkozar: Why change this? Create is there for ages and is idempotent. Deploy should also modify before deploying, so no need to need modify. Also one would expect create to exist, not having to use modify for creation.

We could make create and modify just aliases.

nixpkgs-src.json
```diff
@@ -1,6 +1,6 @@
 {
-    "owner":  "NixOS",
+    "owner":  "input-output-hk",
```
domenkozar: This will break Hydra, since it doesn't account for a different owner in `jobsets/default.nix`

staging-infra.yaml
```diff
@@ -0,0 +1,28 @@
+args:
+  configurationKey:
```
domenkozar: Could we somehow leave these inputs out? They are not relevant for this deployment.

iohk/NixOps.hs
deepfire: Ok, will do.

nixpkgs-src.json
deepfire: This is a temporary stub for testing -- will drop that change soon.

staging-infra.yaml
deepfire: That seems to require some meta-information to be attached to the properties -- which can then be used to filter the generated config.

Do you think that would be worth the complication?

iohk/NixOps.hs
deepfire: Done.

deployments/infrastructure.nix
```diff
@@ -43,44 +43,14 @@ in {
...
-          "/etc/"
-        ];
-      };
-    };
-
     networking.firewall.allowedTCPPortRanges = [
       { from = 24962; to = 25062; }
     ];
   };
+  resources = {
```
domenkozar: This should be part of staging and production.

iohk/iohk-ops.hs
```diff
@@ -313,9 +311,6 @@ runNew o@Options{..} New{..} args = do
   echo $ "-- " <> (unsafeTextToLine $ configFilename) <> " is:"
   cmd o "cat" [configFilename]
```
domenkozar: Create should still exist here, to avoid extra steps when creating a new deployment

deployments/infrastructure-env-staging.nix
```diff
@@ -13,6 +13,10 @@ in {
       ./../modules/datadog.nix
     ];
 
+    ##
```
domenkozar: Let's open an issue for fixing DNS issues and add a TODO here for it.

deployments/infrastructure.nix
domenkozar: cc @deepfire

PR #205, opened 2017-11-22
[DEVOPS-157] Propagate devnet fix

deepfire:
Propagate the `devnet` configuration fix.


deepfire:
@domenkozar, @cleverca22, it would be nice to have your review.

domenkozar:
This is already included in current commit of cardano-sl in `develop`

PR #211, opened 2017-12-12
[DEVOPS-458] fix cardano PR status messages

cleverca22:


cleverca22:
oh, was already fixed in https://github.com/input-output-hk/iohk-ops/commit/966eae7cf1cb7a326a20e2e4d111958acb7daaca but not deployed

PR #216, opened 2017-12-17
increase nginx client limit from 512 to 4096

cleverca22:


cleverca22:
i'm not sure why, but `localhost` began refering to the ipv6 localhost,and explorer is only listening on the ipv4 one

PR #219, opened 2017-12-20
[CSL-1876] Update scripts that are using auxx

hithroc:


domenkozar:
@hithroc do we have the docs for the cli DSL? I don't understand these changes.

hithroc:
No, I don't think so.
cc @int-index

int-index:
We have them for a long time now, `docs/auxx.md` in the Cardano SL repo. Let me know if you need more info there, I will add it.

PR #221, opened 2017-12-22
[DEVOPS-61] Add terraform to bootstrap cache-s3 bucket

domenkozar:
Required for https://github.com/input-output-hk/cardano-sl/pull/2166

### DIFF
--- a/terraform/appveyor-s3-cache/README.md
+++ b/terraform/appveyor-s3-cache/README.md
@@ -37,4 +37,4 @@ environment:
       secure: BSNfEghh/xxx=
     AWS_SECRET_ACCESS_KEY
       secure: BSNfEghh/xxx=
-    BUCKET: XXX
+    S3_BUCKET: XXX
--- a/terraform/appveyor-s3-cache/README.md
+++ b/terraform/appveyor-s3-cache/README.md
@@ -1,0 +1,40 @@
+# Prerequisites
+
+To start with, create a bucket to store terraform state:
+
+- name: iohk-ops-tfstate
+- region: US West (N. California)
+- turn on versioning
+
+Side note: terraform will pick credentials from ~/.aws/credentials and you can switch profiles by doing
+
+     export AWS_PROFILE=iohk
+
+To start and see what terraform will do:
+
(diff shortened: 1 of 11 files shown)


terraform/appveyor-s3-cache/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+And then modify appveyor.yml with the output values:
+
+environment:
+  global:
+    AWS_ACCESS_KEY_ID:
+      secure: BSNfEghh/xxx=
+    AWS_SECRET_ACCESS_KEY
+      secure: BSNfEghh/xxx=
+    BUCKET: XXX
```
lehins: Same as in https://github.com/input-output-hk/cardano-sl/pull/2166/files#r158594166 variable name is `S3_BUCKET`

lehins:
@domenkozar This looks great!

terraform/appveyor-s3-cache/main.tf
```diff
@@ -0,0 +1,31 @@
+provider "aws" {
+  region = "us-west-1"
+}
+
+terraform {
+  backend "s3" {
+    bucket  = "iohk-ops-tfstate"
```
deepfire: @domenkozar, is this snippet with the TF state S3 bucket intended to be copied to all TF deployments, or maybe we could separate it into a separate module, outside of the `appveyor-s3-cache`-specific part?

domenkozar: Yes, that's going to be shared between deployments. We can do that later on, once we expand tf usage

lehins: @domenkozar @deepfire It is common to use the same bucket for all deployments done with TF, although having a __separate__ TF state file is a good practice. So, I'd recommend for each logical deployment to have it's own snippet like that, granted with the same bucket name. Reasoning for this is fairly straightforward, you don't want to have one global state for your whole infrastructure. Of course, this is just a suggestion. :)

domenkozar: @lehins thanks!

PR #222, opened 2017-12-27
[DEVOPS-528] Cardano SL Buildkite migration

deepfire:


nixpkgs-src.json
```diff
@@ -1,6 +1,6 @@
 {
-    "owner":  "NixOS",
+    "owner":  "deepfire",
```
domenkozar: Let's wait for https://github.com/NixOS/nixpkgs/pull/32387 to be in 17.09

PR #234, opened 2018-01-12
[DEVOPS-528]  Two more Buildkite Linux agents

deepfire:


domenkozar:
@deepfire CI fails because nixops state cleanup doesn't seem to work correctly? It seems like an existing deployment is there on the build slave:

```
error: state file contains multiple deployments with the same name, so you should specify one using its UUID
```

deepfire:
@domenkozar, this is due to https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-623

deepfire:
..and the failure happened prior to the debug PR being merged..

PR #236, opened 2018-01-16
[DEVOPS-382][DO-NOT-MERGE]  Enable production infra to be deployed on staging

deepfire:
This is to facilitate infrastructure upgrade testing.


deepfire:
The CI wouldn't work, because it's a branch with an ancient base -- on purpose.

PR #239, opened 2018-01-22
[DEVOPS-623]  ci.sh:  divert $HOME to $(mktemp -d)

deepfire:


scripts/ci.sh
```diff
@@ -53,6 +55,7 @@ cleanup() {
                 test -z "${CLEANUP_CONFIGS}" ||
                         rm -f                ${depl}'.yaml'
         done
+        rm -rf ${homestate}
```
domenkozar: Do we still need manual cleanup? I don't think so.

PR #240, opened 2018-01-24
[DEVOPS-606] Change installer uploading script to use buildkite instead of travis-ci

rvl:


iohk/Buildkite/Pipeline.hs
```diff
@@ -0,0 +1,49 @@
+{-# LANGUAGE OverloadedStrings #-}
```
domenkozar: I don't see PipelineDefinition nor PipelineStep used anywhere (except in tests), am I missing something? :)

iohk/UpdateLogic.hs
```diff
@@ -106,23 +142,27 @@ data VersionJson = VersionJson {
...
-    f2 (Left err) = error $ "unable to parse .travis.yml: " <> err
-    f3 = global $ env (f2 f1)
-  lookup' "VERSION" f3
+extractVersionFromTravis :: TravisYaml -> Maybe T.Text
+extractVersionFromTravis yaml = lookup' "VERSION" (global (env yaml))
+
+-- | Get VERSION environment variable from pipeline definition.
+-- First looks in global env, otherwise finds first build step with a version.
+extractVersionFromBuildkite :: BK.PipelineDefinition -> Maybe T.Text
+extractVersionFromBuildkite BK.PipelineDefinition{..} = getVer plEnv <|> headMay stepVers
```
domenkozar: Most of these parsing functions would go away if we stored versions in a central place in git repos.

We can leave it as is, but once DEVOPS-59 is done this won't be needed anymore.

domenkozar:
@rvl to what extent has this been tested? Otherwise I have no complains except that in the future we should revise how we extract versions in a simpler matter.

rvl:
> to what extend has this been tested?

I have tested:

 - `find-installers` on both buildkite (rev 66c7ef5723ef363d6701133654f4c65848821151)
   and travis (rev a267ae05daaf3947fe679eac56d1257a8775b27b)
 - `set-version-json` on buildkite, up until where it requires access to S3
   (didn't want to run further because that could break things)
 - `s3upload` on buildkite, up until it fails to find
   `/nix/store/hjvv6dxy197c9mjc2gh635am0c0shx5l-cardano-sl-auxx-1.0.3/bin/cardano-hash-installer`
   in my nix store.

Actually that reminds me, I should copy that path into my nix store so
that the full upload process can be tested.


Another thing: is the use of `pkgs.fetchurl` instead of `builtins.fetchurl` ok? (see b6f468d034241fa6d59dc21ce975a1b97f4239b3) The latter provides hash verification.

domenkozar:
cc @cleverca22

rvl:
OK, have tested `s3upload` and `set-version-json` for both Buildkite and Travis revisions using a temporary s3 bucket. Seems to work.

One more thing. Buildkite build numbers are currently lower than travis build numbers so the daedalus version might need to be bumped to avoid filename collisions.

domenkozar:
Ah, that's a good point. Could we somehow change the name just to make it unique?

PR #241, opened 2018-01-25
[DEVOPS-636]  explorer:  now rejects --address and --listen

deepfire:


modules/cardano-service.nix
```diff
@@ -13,8 +13,10 @@ let
 
   command = toString [
     cfg.executable
-    (optionalString (cfg.publicIP != null) "--address ${cfg.publicIP}:${toString cfg.port}")
-    "--listen ${cfg.privateIP}:${toString cfg.port}"
+    (optionalString (cfg.publicIP != null && params.name != "explorer")
```
domenkozar: @akegalj any explanation why?

akegalj: I am not aware of this problem.

@sectore @ksaric do you know what happened and why explorer changed its interface ?

ksaric: Not sure what was that for :man_shrugging: 

@domenkozar Can you explain why you would need that? 
We did refactor the parameters around (several times), seems like this is all that's _explorer_ specific - https://github.com/input-output-hk/cardano-sl/blob/develop/explorer/src/explorer/NodeOptions.hs#L31

domenkozar: > Can you explain why you would need that?

From operational perspective we need to set IP and port where explorer will listen. But it really is a passive node, so we skip that. I just wonder why there's no way to set this or is the intention that it's not useful?

ksaric: > I'd imagine we strive to keep interface uniform.

Yes. That was intentional. Really. I swear. Ok, you got me, I have no clue why that was removed. The original idea was to keep the interface uniform, but as you can see we have some specific params for `explorer` and `wallet`. I suspect this was a case of "Why do we even need that? Delete. Delete.". 
But if you guys require it, you can create an issue and we can add it back :man_shrugging:

domenkozar: I'd imagine explorer has the same options as node + additional ones. Is there a reason why that can't work?

PR #242, opened 2018-01-25
[DEVOPS-607] Disallow non-root set authorized_keys on all deployments

rvl:
Have not been able to test this in a new developer cluster yet.
However there is a good chance it will work because the same config change worked in another NixOS setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,4 +20,4 @@ Collection of tooling and automation to deploy IOHK infrastructure.
 
 ### Getting SSH access
 
-Append https://github.com/input-output-hk/pos-prototype-deployment/blob/master/lib.nix#L46 and submit a PR.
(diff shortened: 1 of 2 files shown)


rvl:
Have tested configs with:

1. Build nixops machines (topology file was cut down to save memory)
```
io --deployer 127.0.0.1 -c mainnet.yaml deploy --no-explorer-rebuild --build-only
```

2. Check resulting configs
```
#!/usr/bin/env bash

set -e

nixops_machines=${1:-/nix/store/0hajskww7mw1x0dqsj3nsg4fahfbxvyb-nixops-machines}
machines="c-a-1 r-a-1 p-b-1"

for machine in $machines; do
    sshd_config=$(awk '/ExecStart=/ { print $3; }' $nixops_machines/$machine/etc/systemd/system/sshd.service)
    echo "$machine: $(grep AuthorizedKeysFile $sshd_config)"
done
```

PR #244, opened 2018-01-29
[DEVOPS-639] Recovery service instance terraform

rvl:
Questions:
 
1) Is S3 statefile storage OK? The S3 key gets namespaced by terraform "workspaces".
2) Are terraform "workspaces" acceptable?
3) Should we remove AWS credentials from the variables file and just use environment variables (e.g. `AWS_PROFILE`). `AWS_PROFILE` already needs to be used for the S3 backend because terraform doesn't support variable interpolation in backend configs.
4) Should this be integrated with the `iohk-ops` tool?
5) What to do with DNS records?


domenkozar:
1/2) Yes, this should suffice for now.
3) Yes, these shouldn't be part of terraform, although getting AWS_PROFILE to work can be tricky. Can you look into this and write up docs how it would be used instead?
4) At some point, but we have to first establish "lessons learned" to agree how to handle terraform deployments as a team before we automate it. My 2c :)
5) Document how it should be provisioned via cloudflare - I can help you with how it's done

jmitchell:
@domenkozar, @rvl and I had a chat and decided that due to the 3rd party support requirements, it would be best to revert to managing credentials `terraform.tfvars`. We also concluded that due to the unique security constraints (separate AWS accounts per staging/production) and, again, supporting the 3rd party, we won't be using an S3 bucket to manage the state file. IIRC, we determined that workspaces aren't necessary under this simplified model, assuming the staging and production are deployed from different contexts with distinct credentials and `terraform.tfvars` files.

PR #249, opened 2018-01-30
[CSL-2186] Use cardan-sl logging config

domenkozar:


domenkozar:
Needs cardano-sl bump first.

PR #251, opened 2018-01-30
allow root to use ~/.ssh/authorized_keys

cleverca22:
nixops doesnt add its own keys to the nixos config, and assumes that the `~/.ssh/authorized_keys` setup by aws will remain in effect

### DIFF
--- a/modules/common.nix
+++ b/modules/common.nix
@@ -1,4 +1,4 @@
-{ config, pkgs, ... }:
+{ config, pkgs, lib, ... }:
 
 with (import ./../lib.nix);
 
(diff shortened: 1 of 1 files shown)


modules/common.nix
```diff
@@ -20,6 +20,10 @@ in {
   # Non-root users are not allowed to install authorized keys.
   services.openssh.authorizedKeysFiles = pkgs.lib.mkForce
     [ "/etc/ssh/authorized_keys.d/%u" ];
+  services.openssh.extraConfig = lib.mkOrder 999 ''
```
domenkozar: @cleverca22 can't we just put `/root/.ssh/authorized_keys` into services.openssh.authorizedKeysFiles?

deepfire: @domenkozar, that was the first question I asked Michael yesterday -- and he suggested that the option is effective for _every_ user, which is somewhat conceptually ugly.

deepfire:
I've tested this commit in the scope of provisional jumpserver deployment:
1. the problem appears without this commit (can't log onto any newly deployed machine)
2. cherry-picking the commit fixes the problem

PR #258, opened 2018-02-11
[DEVOPS-666] find-installers: fix logic for getting cardano revision

rvl:
It gets the cardano revision from the buildkite cardano build.

Due to my mistake in #240 (DEVOPS-606), it was getting the most recent buildkite cardano build, rather than the one referred to in the build logs.


deepfire:
Looks good to me.

PR #264, opened 2018-02-19
[DEVOPS-683] Add zendesk token to report-server

rvl:
Works in a development cluster.

### DIFF
--- a/modules/report-server.nix
+++ b/modules/report-server.nix
@@ -1,6 +1,6 @@
 with (import ./../lib.nix);
 
-params:
+params: environment:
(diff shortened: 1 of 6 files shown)


modules/report-server.nix
```diff
@@ -26,6 +26,32 @@ in {
         type = types.package;
         default = (import ./../default.nix {}).cardano-report-server-static;
       };
+      zendesk = {
+        email = mkOption {
+          type = types.string;
```
cleverca22: https://github.com/NixOS/nixpkgs/blob/master/lib/types.nix#L203-L205

PR #266, opened 2018-02-20
[DEVOPS-382] staging infra master

deepfire:
This is a sister PR to https://github.com/input-output-hk/iohk-ops/pull/191

### DIFF
--- a/deployments/infrastructure.nix
+++ b/deployments/infrastructure.nix
@@ -16,7 +16,6 @@ let
     ];
   };
 in {
-  require = [ ./ugly-fix.nix ];
   hydra = { config, pkgs, ... }: {
     # On first setup:
 
--- a/deployments/infrastructure.nix
+++ b/deployments/infrastructure.nix
@@ -46,10 +46,4 @@ in {
       { from = 24962; to = 25062; }
     ];
   };
-  resources = {
-    datadogMonitors = (with (import ./../modules/datadog-monitors.nix); {
-      disk = mkMonitor disk_monitor;
-      ntp = mkMonitor ntp_monitor;
-    });
-  };
 }
--- a/deployments/infrastructure.nix
+++ b/deployments/infrastructure.nix
@@ -43,44 +43,14 @@ in {
 
     environment.systemPackages = [ iohk-pkgs.iohk-ops ];
 
-    users = {
-      users.staging = {
-        description     = "cardano staging";
-        group           = "staging";
-        createHome      = true;
-        isNormalUser = true;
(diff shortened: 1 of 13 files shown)


iohk/iohk-ops.hs
```diff
@@ -165,8 +164,8 @@ centralCommandParser =
      --                           (Nixops
      --                            <$> (NixopsCmd <$> argText "CMD" "Nixops command to invoke")
      --                            <*> ???)) -- should we switch to optparse-applicative?
-     ("create",                 "Create the whole cluster",                                         pure Create)
-   , ("modify",                 "Update cluster state with the nix expression changes",             pure Modify)
+     ("modify",                 "Update cluster state with the nix expression changes",             pure Modify)
+   , ("create",                 "Same as modify",                                                   pure Modify)
```
jmitchell: Should we just remove `create`? I see the internal change happened at 605a4e4.

deployments/infrastructure.nix
```diff
@@ -43,44 +42,14 @@ in {
...
-          "/etc/"
-        ];
-      };
-    };
-
     networking.firewall.allowedTCPPortRanges = [
       { from = 24962; to = 25062; }
     ];
   };
+  resources = {
```
domenkozar: This should be in `deployments/infrastructure-target-aws.nix`

iohk/iohk-ops.hs
```diff
@@ -313,9 +311,6 @@ runNew o@Options{..} New{..} args = do
   echo $ "-- " <> (unsafeTextToLine $ configFilename) <> " is:"
   cmd o "cat" [configFilename]
```
domenkozar: There should be a call to create here. The idea is to make sure it's either created or modified. We also want to bring create back to backwards compatibitily

iohk/iohk-ops.hs
deepfire: Right, missed that..

deepfire: Ah, I didn't miss that, actually, it's an alias to modify, precisely for backward compat -- already forgot this..

iohk/iohk-ops.hs
deepfire: @domenkozar, `io deploy` runs `io modify` (which should be probably renamed into `update`), which runs either `nixops modify` or `nixops create`, as appropriate.

Also, `io create` is still there and it runs `io modify`.

deployments/infrastructure.nix
deepfire: Done..

PR #267, opened 2018-02-21
[DEVOPS-683] report-server: Pass on logs to Zendesk by default

rvl:
Turns out the `--zd-send-logs` option is wanted.


rvl:
Thanks

PR #273, opened 2018-02-23
[DEVOPS-643] Install script for buildkite-agent on nix-darwin

rvl:


scripts/deploy-buildkite-darwin/README.md
```diff
@@ -0,0 +1,87 @@
...
+The `buildkite-agent` package is actually plucked from a revision on
+the `nixos-unstable` branch because there is no build for Mac in
+17.09.
+
+## After that
+
+Set up the `/Users/admin/buildkite` directory with the necessary
+secrets as follows.
+
+1. The following variables are required in `buildkite_aws_creds` for
```
domenkozar: I'd specify an absolute path here: `/Users/admin/buildkite/buildkite_aws_creds`

scripts/deploy-buildkite-darwin/README.md
```diff
@@ -0,0 +1,87 @@
...
+   artifact uploads to work:
+
+       ```
+       export BUILDKITE_S3_ACCESS_KEY_ID=AK...
+       export BUILDKITE_S3_SECRET_ACCESS_KEY=...
+       export BUILDKITE_S3_DEFAULT_REGION=...
+       ```
+
+2. Put the agent token into `buildkite_token`.
```
domenkozar: Same here ^^

scripts/deploy-buildkite-darwin/buildkite.hs
```diff
@@ -0,0 +1,214 @@
...
+  bkDir <- (</> "buildkite") <$> home
+  dataPath <- liftIO getDataPath
+  setupConfiguration (dataPath </> "darwin-configuration.nix") bkDir nixPath
+  setupBuildkiteDir bkDir
+  darwinRebuild
+  secureBuildkiteDir bkDir
+
+  -- enable xcode command line developer tools because some things in
+  -- the npm build require it.
+  enableXCodeCL
```
domenkozar: Huh, `npm build` of daedalus? I don't think we have xcode installed on any of the existing macos so I'm surprised about this.

scripts/deploy-buildkite-darwin/darwin-configuration.nix
```diff
@@ -0,0 +1,124 @@
...
+
+  # Used for backwards compatibility, please read the changelog before changing.
+  # $ darwin-rebuild changelog
+  system.stateVersion = 2;
+
+  # You should generally set this to the total number of logical cores in your system.
+  # $ sysctl -n hw.ncpu
+  nix.maxJobs = 4;
+  nix.buildCores = 1;
+  nix.useSandbox = false;  # this seems to break things when enabled
```
domenkozar: Yes, this can be enabled after Nix 2.0 and careful testing :)

scripts/deploy-buildkite-darwin/darwin-configuration.nix
```diff
@@ -0,0 +1,124 @@
...
+      ln -s ${prog} $out/bin
+    '';
+  in {
+    enable = true;
+    package = unstablePkgs.buildkite-agent;
+    runtimePackages = with pkgs; [ bash nix git ] ++
+      map symlinkBin [ "/usr/sbin/pkgutil" ] ++
+      # These are required for daedalus build.
+      # fixme: update daedalus pipeline scripts so that it declares all of its dependencies
+      [ python27 gnumake curl nodejs-6_x nodePackages.node-gyp nodePackages.node-pre-gyp electron ];
```
domenkozar: This shouldn't be here and it should belong to daedalus as noted - I think we should fix it before merging this.

scripts/deploy-buildkite-darwin/nuke-nix.hs
```diff
@@ -0,0 +1,52 @@
+#! /usr/bin/env nix-shell
```
domenkozar: Is this used somewhere?

rvl:
> we should make this the default setup for macOS machines and move it to top-folder like nix-darwin

OK moved.

> we also setup hydra slaves for macOS and to support that initially we could just allow this setup not to perform buildkite parts if told so (we can later add slave setup)

OK, will work on that shortly. Should just be a matter of splitting the config and using imports.

> is deploy-buildkite idempotent? it's good to note if that should be run each time we update a macOS machine or just nix-darwin rebuild

Yes it's idempotent. If we want, we could use the script to update deployments. I have updated the readme to explain.

> I'd specify an absolute path here: `/Users/admin/buildkite/buildkite_aws_creds`

OK fixed.

> Huh, npm build of daedalus? I don't think we have xcode installed on any of the existing macos so I'm surprised about this.

Turns out this wasn't required. I was tricked by a spurious message in the npm build output. I have removed it.

> This shouldn't be here and it should belong to daedalus as noted - I think we should fix it before merging this.

OK, have removed and made the pull request input-output-hk/daedalus#746.

> `nuke-nix.hs`: Is this used somewhere?

It's a "development tool". Useful to see if the deploy script works on a fresh install. I have added a comment to the top of the file.

nix-darwin/darwin-configuration.nix
```diff
@@ -0,0 +1,72 @@
...
+    gc-keep-derivations = true
+    gc-keep-outputs = true
+  '';
+
+  nix.binaryCachePublicKeys = [ "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=" ];
+  nix.binaryCaches = [ "https://hydra.iohk.io" ];
+  nix.trustedUsers = [ "@admin" ];
+
+  nix.nixPath = [
+    "nixpkgs=https://nixos.org/channels/nixpkgs-unstable/nixexprs.tar.xz"
```
domenkozar: would be nice if we could pin this to `unstablePkgs` commit as well

rvl: Maybe this is a bit confusing and magic... the values in this file actually get sedded by the deploy script. I could remove these initial values to make it clearer?

The deploy script pins to the 17.09 channel but not a specific commit. Is that ok?

domenkozar: if it's 17.09 then we can just use `fetch-nixpkgs.nix` in the root folder :)

Do we need to sed it? Can't we just put those here?

rvl: The nix path is defined in haskell because it's needed before running the nix-darwin installer. I have edited the nix to make it clear that the values are templated.

rvl:
What's left:
 * sort out the upstream PR (LnL7/nix-darwin#70) -- needs a simple test case
 * update the nix-darwin rev
 * let the new buildkite agent run some builds and check that it continues to work (cardano-sl and daedalus builds should work)
 * squash and merge
 * run deploy script to convert existing macs to nix-darwin
 * add nix module for hydra build slave and deploy (will probably do this in another ticket)

rvl:
DE309, the little buildkite-agent that could, seems to be going alright, although a little slow.

I have updated the `nix.buildCores` setting.

PR #280, opened 2018-03-06
[DEVOPS-711] find-installers: Disable downloading from travis

rvl:
`daedalus/.travis.yml` was wrong. Since the failing download exits the program, just ignore travis and use buildkite instead.


deepfire:
@domenkozar, policy question -- what branch should the base be?

This is sort-of a hotfix, to enable the proposal to be made with the tooling, so this suggests `master` as the base.

However, this is also a move to deprecate Travis, which is no laughing matter..

rvl:
See also: #256

domenkozar:
Since buildkite proved to work - my vote would be to ship this.

domenkozar:
(we can always hotfix to get travis back in)

PR #281, opened 2018-03-06
re-arrange shell.nix to add new features and improve perf

cleverca22:
plain `nix-shell` gives you a shell with just `io`, `nixops`,and `terraform` and no cardano or ghc
`nix-shell -A ioSelfBuild` gives you a shell with `io` and `ghcid-io` functions and a ghc that can build iohk-ops
`nix-shell -A withAuxx` gives a shell with io, auxx, and tools, no ghc

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -65,6 +65,10 @@ GENERAL_OPTIONS="--verbose --deployer 0.0.0.0"
 COMMON_OPTIONS="--topology topology-min.yaml"
 CARDANO_COMPONENTS="Nodes ${WITH_EXPLORER:+Explorer} ${WITH_REPORT_SERVER:+ReportServer}"
 
+nix-build default.nix -A cardano-sl-tools -o cardano-sl-tools
+
+export PATH=$PATH:./cardano-sl-tools/bin/
+
 if test -n "${WITH_STAGING}"; then
 CLEANUP_DEPLS="${CLEANUP_DEPLS} test-stag"
 ${IOHK_OPS}               new  --config 'test-stag.yaml'   --environment staging    ${COMMON_OPTIONS} 'test-stag'    ${CARDANO_COMPONENTS}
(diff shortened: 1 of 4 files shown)


shell.nix
```diff
@@ -4,15 +4,29 @@
...
+    shellHook = ''
+      function io {
+        runhaskell -iiohk iohk/iohk-ops.hs
+      }
+      function ghcid-io {
+        ${pkgs.haskellPackages.ghcid}/bin/ghcid -c "ghci -iiohk iohk/iohk-ops.hs"
+      }
+    '';
+  });
+  withAuxx = pkgs.runCommand "shell" {
```
deepfire: @cleverca22, so this is to be used like:
```
nix-shell -A withAuxx
```
right?

cleverca22: yep

scripts/ci.sh
```diff
@@ -65,6 +65,10 @@ GENERAL_OPTIONS="--verbose --deployer 0.0.0.0"
 COMMON_OPTIONS="--topology topology-min.yaml"
 CARDANO_COMPONENTS="Nodes ${WITH_EXPLORER:+Explorer} ${WITH_REPORT_SERVER:+ReportServer}"
 
+nix-build default.nix -A cardano-sl-tools -o cardano-sl-tools
```
domenkozar: Hmm, why is this needed? Shouldn't the code just use nix-shell if it needs these tools?

cleverca22: the CI runs `iohk-ops` without a nix-shell, and was relying on the fact that it was `makeWrapper'`d to have cardano-tools in `PATH`

and the whole point of this PR is to not depend on cardano when building io

domenkozar: When does CI need it exactly?

domenkozar:
OK, this just needs a docs update then: https://github.com/input-output-hk/iohk-ops#development

PR #278, opened 2018-03-02
Master to develop: a summary update

deepfire:

PR #284, opened 2018-03-07
Develop proposed merge

deepfire:
Supercedes https://github.com/input-output-hk/iohk-ops/pull/278


rvl:
Not sure why there are two merge commits e19212f68e36afa00a690bde4329ad0d326bb08c and 2224a07276a5b9f6fe142d56f9f344fc3d764016 rather than just one?

Could the conflict between `terraform` and `terraform_0_11` be resolved in favour of `terraform_0_11`? This program is picky about versions and the state files aren't backwards compatible. So although we use stable nixpkgs revision, I still want to set the version to avoid future surprises.

rvl:
Doh I noticed that the `terraform` variable is actually referring to `terraform_0_11.withPlugins...`. So it's OK the merge is all good. :blush:

PR #285, opened 2018-03-08
[DEVOPS-697] add the new benchmarking deployment file

cleverca22:
this allows spawning an array of wallets, with a customized topologyfile&starttime and using them to benchmark a relay

### DIFF
--- a/cardano-sl-src.json
+++ b/cardano-sl-src.json
@@ -1,6 +1,6 @@
 {
     "url": "https://github.com/input-output-hk/cardano-sl",
     "fetchSubmodules": "true",
-    "sha256": "1dj9ps1f5k5pz8w62j514cqmssnp6a69hlnq4n16v3svbshimfkf",
-    "rev": "7e0731dbfdf967ecc12ad996d8dd86caeed0738c"
+    "sha256": "1s5v2abfkyyv8n4mdyj029kpng4pbbkdl2d19wwagcmyl1zc3br0",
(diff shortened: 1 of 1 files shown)


domenkozar:
Also needs an issue id :)

modules/cardano-benchmark.nix
```diff
@@ -0,0 +1,89 @@
...
+      systemStart = mkOption {
+        type = types.bool;
+        default = true;
+      };
+    };
+  };
+
+  config = {
+
+    let
```
domenkozar: Don't think this will actually evaluate?

cardano-sl-src.json
```diff
@@ -1,6 +1,6 @@
 {
     "url": "https://github.com/input-output-hk/cardano-sl",
     "fetchSubmodules": "true",
-    "sha256": "110dqfpj1yfvz70zrs479y050pr2vf973ddgsfghzm5pvxz0chd3",
-    "rev": "c3ba8aa37f8b0d632222115fd10a409e1396feaf"
-}
\ No newline at end of file
+    "sha256": "0hrhzk6p0h3bk0zf7shqwrr44683vdqa509ynfi6lzxivhii5lvm",
+    "rev": "6d83230489ff70985ffd6b7c785286d7535566c4"
```
disassembler: This rev probably needs updated. The changes needed for launch script overrides are now in develop.

PR #294, opened 2018-03-13
[DEVOPS-674] buildkite darwin: run git clean before build

rvl:
I would like to do the same change for the linux buildkite-agents, but the `services.buildkite-agent.extraConfig` option isn't implemented in 17.09.

This is an initial step towards more build purity and will help in the short-term with disk space shortages.


domenkozar:
my 2c: I'd add this to daedalus repo since that one is polluting disk space, meanwhile DEVOPS-679 as a proper fix.

rvl:
OK, sure.

PR #298, opened 2018-03-16
fix hydra

cleverca22:


cleverca22:
https://github.com/NixOS/hydra/pull/547
fix has also been upstreamed

PR #303, opened 2018-03-20
[DEVOPS-678] shellcheck for iohk-ops shell scripts

disassembler:


domenkozar:
@disassembler one error left is seems :)

scripts/shellcheck.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+EXIT_STATUS=0
+
+#for i in $(find . -name '*.sh')
```
jmitchell: Let's either do it this way, or remove the comment.

.buildkite/pipeline.yml
```diff
@@ -7,3 +7,11 @@ steps:
       NIX_REMOTE: daemon
     agents:
       system: x86_64-linux
+  - label: 'iohk-ops-shellcheck'
+    command: 'export NIX_PATH="nixpkgs=$(/run/current-system/sw/bin/nix-build "$(dirname -- $(readlink -f -- ${BASH_SOURCE[0]}))/fetch-nixpkgs.nix" -o nixpkgs)"; export NIX_BUILD_SHELL=/run/current-system/sw/bin/bash; /run/current-system/sw/bin/nix-shell -p shellcheck -p coreutils --run scripts/shellcheck.sh'
```
deepfire: Do you think it's worth factoring this?

There's actually a script in `cardano-sl` that implements something very close to what's needed: https://github.com/input-output-hk/cardano-sl/blob/develop/scripts/ci/nix-shell.sh

PR #307, opened 2018-03-28
[DEVOPS-656] Update proposal script

rvl:
- The script is implemented as an `io update-proposal` subcommand with
  three further subcommands:

  - `init` -- sets up the work directory and blank config file
  - `generate` -- copies keys, sets up database, downloads and checks installers.
  - `submit` -- sends transaction to a relay, the ip address being
    passed on the command line.

- The script sets up work directories which have the network name and date in their path name.

- Inside the work directory is a yaml config file, and subdirectories for logs, keys, database, installers.

- Need some feedback about this -- have I made it better or worse.

- Automatic testing of turtle scripts (shell scripts in general) is
  tricky. Maybe we need to brainstorm this.


domenkozar:
we also need to reflect current documentation according to these changes.

iohk/iohk-ops.hs
```diff
@@ -178,6 +180,7 @@ centralCommandParser =
    , ("fromscratch",            "Destroy, Delete, Create, Deploy",                                  pure FromScratch)
    , ("reallocate-core-ips",    "Destroy elastic IPs corresponding to the nodes listed and redeploy cluster",
                                                                                                     pure ReallocateCoreIPs)
+   , ("update-proposal",        "Fixme command help",                                               UpdateProposal <$> parseUpdateProposalCommand)
```
jmitchell: "Fixme"

rvl:
Update procedure page on the wiki is updated with instructions for this.

rvl:
Have updated and finalized the docs after another test run.

Was having issues running `cardano-auxx` from the develop branch. I will investigate that separately and possibly send a bug report to cardano-sl.

PR #309, opened 2018-03-29
[PB-51] Added option for policy files and output parameters of nix scripts

365andreas:
Policy files will change in `modules/cardano-sevice.nix` file.
Parameters of nix scripts  are recorded in `run-bench-args.txt` and `collect-data-args.txt`


365andreas:
No as we are not calling it inside nix-scripts (run-bench, collect-data). I think that devops will know how to automate it in the best way.

PR #311, opened 2018-04-02
[DEVOPS-758]  cardano-sl:  bump to tip of release/1.1.1

deepfire:


rvl:
I'm guessing this is not needed any more?

PR #317, opened 2018-04-11
[DEVOPS-787] iohk-ops: Update find-installers for revision pinning

rvl:
The cardano-sl git revision is now kept in a JSON file. There's no more need to search Buildkite logs.

PR #320, opened 2018-04-12
[DEVOPS-710] iohk-ops update-proposal: add installer signatures

rvl:
This is the scripting for running gpg and uploading the signatures to S3.
@deepfire is working on signing key management.
It's based off PR #317, so that needs to be merged first.

### DIFF
--- a/iohk/UpdateProposal.hs
+++ b/iohk/UpdateProposal.hs
@@ -22,6 +22,7 @@ import qualified Data.HashMap.Strict as HM
 import qualified Control.Foldl as Fold
 import Data.Char (isHexDigit)
 import Data.Bifunctor (first)
+import Data.Maybe (fromMaybe, catMaybes)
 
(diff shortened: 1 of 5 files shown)


iohk/UpdateProposal.hs
```diff
@@ -428,6 +442,8 @@ updateProposalFindInstallers opts env = do
   res' <- moveInstallersToWorkDir opts res
   writeWikiRecord opts res'
   storeParams opts (UpdateProposalConfig2 params res')
+  echo "*** Installers are now in the work dir. Optionally sign them with:"
```
disassembler: It looks like we're signing them on line 490, so why are we telling the user to sign them manually?

rvl:
Fixed the log message. I didn't test this on the deployer, only locally with my own gpg key and the update-system-testing s3 bucket.

PR #326, opened 2018-04-20
[DEVOPS-806] Update find-installers version scheme

rvl:
* Versions and commit ids are now found by using nix.

* It now supports downloading of both staging and mainnet installers (DEVOPS-690).

* iohk-ops find-installers now prints a handy summary when finished ([example](https://github.com/input-output-hk/internal-documentation/wiki/iohk-ops-reference#find-installers)).

* Actually downloading the files is now optional in iohk-ops find-installers.

* Moved s3upload command into subcommand of update-proposal.

* It should be fairly simple now to add Linux installer support (DEVOPS-13).

* Straightened out the data types a little bit.

* Split out the code related to getting Daedalus installer version info into its own module.

* Some churn with converting some values to FilePath type.

### DIFF
--- a/iohk/UpdateProposal.hs
+++ b/iohk/UpdateProposal.hs
@@ -14,7 +14,7 @@ import Data.Time.Format (formatTime, iso8601DateFormat, defaultTimeLocale)
 import qualified Data.Text as T
 import qualified Data.Text.IO as TIO
 import qualified Data.ByteString.Lazy.Char8   as L8
-import Control.Monad (forM_)
+import Control.Monad (forM_, forM)
 import Filesystem.Path.CurrentOS (encodeString)
 import Data.Yaml (decodeFileEither, encodeFile)
 import Data.Aeson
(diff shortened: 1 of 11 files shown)


iohk/UpdateProposal.hs
```diff
@@ -437,36 +427,36 @@ updateProposalFindInstallers opts env = do
...
+      pure (arch, res { ciResultLocalPath = p })
+    move :: FilePath -> Shell FilePath
     move src = do
-      let src' = fromText src
-      let dst = installersPath opts (filename src')
-      mv src' dst
-      pure (format fp dst)
+      let dst = installersPath opts (filename src)
+      mv src dst
+      void $ chmod writable dst
```
cleverca22: 1: this fails to chmod the files because they are in the nix store
2: the installers shouldn't be writable
3: by moving them, nix will no longer know where they are, and garbage-collection can eat the installers, the proper directory should just be passed directly to `realFindInstallers`

rvl: Thanks, fixed. Also pushed some error message fixes.

iohk/UpdateProposal.hs
```diff
@@ -437,36 +434,18 @@ updateProposalFindInstallers opts env = do
...
+  echo "*** Finished."
+  writeWikiRecord opts res
+  storeParams opts (UpdateProposalConfig2 params res)
+
+installerForEnv :: Environment -> CIResult -> Bool
+installerForEnv env = matchNet . installerNetwork . ciResultLocalPath
+  where matchNet n = case env of
+          Production -> n == Just InstallerMainnet
+          Staging    -> n == Just InstallerStaging
+          _          -> True
```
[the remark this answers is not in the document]
rvl: Thanks, I have added a comment to try and clarify this function.

I think it's OK because `env` comes from the `--config` option. So if `mainnet.yaml` then `env` is `Production`, if `staging-testnet.yaml` then `env` is `Staging`.

So what it's saying is that when running under the production config, proposing an update for the mainnet, it will only find installers containing "mainnet" in their filename. When running under a development config, any installer will do, just so that it's possible to test this script on a development network.

There is a fourth option of the `Environment` type which is `Any` and I think this should be removed at some point, but I didn't do it today.

rvl:
Squashed fixups and rebased on latest develop.

PR #328, opened 2018-04-25
[DEVOPS-710] authority.sh: manage master/subkeys & facilitate signing/verifictiona

deepfire:
*TODO*:
- [x] make shellcheck pass
- [~] possibly take care of USB stick mounting for the master key -- deferred to documentation
- [x] improve usage documentation

Test with:
```
$ authority.sh --dry --selftest
# ..or, if aiming for 100% authentic end-user experience (entering the passphrase..):
$ authority.sh --selftest
```

### DIFF
--- a/doc/IOHK GPG Authority Usage Guidelines.md
+++ b/doc/IOHK GPG Authority Usage Guidelines.md
@@ -1,0 +1,151 @@
+# IOHK GPG signing authority usage guidelines
+
+# Design goals
+
+Establish a GPG authority, to facilitate publicly-verifiable attribution of
+artifacts distributed by IOHK. Since compromise of such an authority entails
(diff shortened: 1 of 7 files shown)


deepfire:
1. The signing authority management document has been finalized.
2. The signing authority has been generated.
3. End-user signature verification document was finalized, tested on all three OSes and sent to translation.

deepfire:
Dropped the explicit, fingerprint-based specification of the GPG key used for installer signing.

I still think it's a mistake -- the key is important enough to warrant a guard against accidental changes.

deepfire:
Having configuration implicit in the state is something we always tried to avoid so far, so it's surprising we're explicitly moving away from that.

deepfire:
We don't appear to need this anymore.

PR #331, opened 2018-04-30
[DEVOPS-777] infra: enable docker service on linux buildkite-agents

rvl:
Required to support mantis docker builds.


rvl:
Thanks!

PR #333, opened 2018-05-07
[DEVOPS-825] Use predefined IAM role for report server

deepfire:
- Accomodations for the S3 access scheme changes.
- Zendesk token spec fix
- ci.sh memory requrements
- find-all-revisions.sh exit status fix

### DIFF
--- a/cardano-sl-src.json
+++ b/cardano-sl-src.json
@@ -1,6 +1,6 @@
 {
   "url": "https://github.com/input-output-hk/cardano-sl",
-  "rev": "f4af8443ef61fa8792f4887f8c58f0e96c9e1f1e",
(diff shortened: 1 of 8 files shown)


deepfire:
@rvl, why?

The deployment is governed by the YAML, of course.

rvl:
OK thanks for explaining that the yaml is generated from `Constants.hs`.
We are hardly going to be running `iohk-ops new` for a mainnet or staging. If we do, the hardcoded role IDs will most likely be wrong anyway. To reduce confusion, they could be set to an obvious placeholder value.

jmitchell:
CI is failing with `value is a string while a Boolean was expected` and it cites `report-server.nix`. I believe the trouble is `tokenFile` is being used as a bool when its type is `types.nullOr types.path`.

PR #335, opened 2018-05-10
[DEVOPS-695] nix-darwin config for hydra build slaves

rvl:
Declarative configuration of macOS Hydra build slaves.


lib.nix
```diff
@@ -77,4 +77,13 @@ in lib // (rec {
     philippKey
     vasilisKey
   ]);
+
+  buildSlaveKeys = with ssh-keys; {
+    macos = devOpsKeys ++ [
+      hydraBuildFarmKey
+      rodneyRemoteBuildKey
```
cleverca22: `rodneyRemoteBuildKey` for testing?

modules/hydra-slave.nix
```diff
@@ -16,7 +16,7 @@ in
 
   environment.systemPackages = [ iohk-pkgs.iohk-ops ];
 
-  users.extraUsers.root.openssh.authorizedKeys.keys = pkgs.lib.singleton ''
-    command="nice -n20 ${pkgs.utillinux}/bin/flock -s /var/lock/lab nix-store --serve --write" ${pkgs.lib.readFile ./../static/id_buildfarm.pub}
-  '';
+  users.extraUsers.root.openssh.authorizedKeys.keys = map (key: ''
+    command="nice -n20 ${pkgs.utillinux}/bin/flock -s /var/lock/lab nix-store --serve --write" ${key}
```
cleverca22: i think the flock part should go, its forcing the slave to only run 1 job at once, and hydra is configured to run 4 at once on each slave

nix-darwin/roles/buildkite-agent.nix
```diff
@@ -8,6 +8,6 @@
   ];
 
   services = {
-    dd-agent.tags = [" group:buildkite-agents"];
+    dd-agent.tags = [" group:buildkite-agents" "group:macos"];
```
cleverca22: we should be able to remove the redundant space before group, that was to fix a buildkite module bug

nix-darwin/roles/hydra-slave.nix
```diff
@@ -0,0 +1,13 @@
...
+
+{
+  imports = [
+    ../modules/basics.nix
+    ../modules/datadog.nix
+    ../modules/hydra-slave.nix
+  ];
+
+  services = {
+   dd-agent.tags = [" group:hydra-and-slaves" "group:macos"];
```
cleverca22: redundant space again in the string

lib.nix
rvl: This is because in my config I want to use a separate ssh key for nix remote builders. Is this OK?

rvl:
Added review fixes and updated readme.

rvl:
Rebased and squashed on develop which should fix the buildkite CI status.

PR #336, opened 2018-05-11
fix the missing /etc/protocols

cleverca22:


rvl:
Looks like a good change. Could you please explain it a bit?

KtorZ:
@rvl See https://github.com/input-output-hk/cardano-sl/pull/2802 and the error we get from the CI at the moment: https://hydra.iohk.io/build/122207/nixlog/3

rvl:
Yes that is one part of the explanation. The other part is to explain the solution. Why `nix.sandboxPaths` works now and didn't before.

KtorZ:
That's a good question. We started to encounter this issue only with #2802 which introduce http**s** in the integration tests. We use to run all testing with the wallet backend in debug mode (i.e. no TLS support) and do all testing request in plain http. 

Why this now fails to identify `tcp` as a valid protocol, I have no clue. Note that we (neither I or Matt Parsons) get the issue locally. I am trusting @cleverca22 expertise & analyze on that one.

cleverca22:
i had forgotten about this option in nix.conf when i helped sam with patching nix directly for the previous issue

fixing it in the config is going to be a lot better then patching nix for every single file

rvl:
Cool thanks

rvl:
I have already made a PR for the `find-all-revisions.sh` script in #338

jmitchell:
@cleverca22, since this is apparently blocking successfully Hydra builds for https://github.com/input-output-hk/cardano-sl/pull/2802 and requires a deployment it's especially important to track this in a YouTrack issue.

Can you create one, put it in the appropriate deploy column, and reference its ID in the title?

PR #343, opened 2018-05-18
Shellcheck hydra

disassembler:


rvl:
Should this be on the `master` branch?
I already made a `fetch-nixpkgs.nix` fix in #435.

PR #345, opened 2018-05-24
Faucet

boothead:
Initial stab at faucet module

### DIFF
--- a/deployments/faucet-env-staging.nix
+++ b/deployments/faucet-env-staging.nix
@@ -1,0 +1,15 @@
+{ globals, ... }: with (import ./../lib.nix);
+
+let params = globals.fullMap.faucet;
+in
+{
+  faucet = { config, resources, ...}: {
+    imports = [
+      ./../modules/development.nix
+    ];
+    services.faucet.faucet-config = {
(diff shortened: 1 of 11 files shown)


staging-testnet.yaml
```diff
@@ -1,37 +1,34 @@
 args:
```
deepfire: I believe we should keep `staging-testnet.yaml` as is -- or even, rename it to `staging.yaml`.

The new file should be named `testnet.yaml`.

Yes, it was terribly confusing.. : -/

deployments/faucet-env-staging.nix
```diff
@@ -0,0 +1,17 @@
+{ globals, ... }: with (import ./../lib.nix);
```
deepfire: Ben, I also think we should create a new kind of environment -- so far we have:
- `development`, for dev time
- `staging` and `mainnet` for corresponding clusters managed by DevOps

The new environment would be called correspondingly -- `testnet`.

..so effectively, this would add `deployments/faucet-env-testnet.nix`, instead.

rvl:
I have included all of these commits in #346 then based my work off that.

PR #346, opened 2018-05-24
[DEVOPS-398] Public testnet deployment

rvl:
This is for the new testnet-deployer host and a module for cardano-sl-faucet.


deployments/testnet-deployer.nix
```diff
@@ -0,0 +1,86 @@
+{ ... }:
```
deepfire: Great that we now finally have the deployer separate!

lib/faucet.html
```diff
@@ -0,0 +1,182 @@
+<!doctype html>
```
deepfire: Does this belong to a separate repository?

deployments/report-server-env-testnet.nix
```diff
@@ -0,0 +1,20 @@
+{ globals, ... }: with (import ./../lib.nix);
```
disassembler: is this needed considering zendesk only works with mainnet?

deployments/security-groups.nix
```diff
@@ -119,6 +120,15 @@ with import ../lib.nix;
                 sourceIp = "0.0.0.0/0";
               }];
             };
+            "allow-to-faucet-${region}" = {
+              inherit region accessKeyId;
+              description = "Access Cardano Faucet";
+              rules = [{
+                protocol = "tcp";
+                fromPort = 80; toPort = 80;
```
disassembler: should be 443/ssl instead of allowing insecure 80 traffic

lib.nix
```diff
@@ -50,44 +50,19 @@ in lib // (rec {
...
-
-  devOpsKeys = with ssh-keys; [
-    jakeKey
-    jakeKey2
-    kosergeKey
-    michaelKey
-    rodneyKey
-    samKey
-  ];
+  inherit devOps csl-developers;
```
disassembler: love this abstraction!

lib/faucet.html
disassembler: I agree, I think we should have a faucet-ui repo (or maybe a derivation on cardano-sl, since that's where the main faucet code is) or something instead of embedding html/vuejs code in iohk-ops.

modules/cardano-faucet.nix
```diff
@@ -0,0 +1,55 @@
...
+
+let
+  recaptcha = {
+    siteKey = lib.fileContents ../static/recaptcha_site_key;
+    secretKey = lib.fileContents ../static/recaptcha_secret_key;
+  };
+  explorerURL = "http://cardano-explorer.cardano-testnet.iohkdev.io/";
+  faucetFrontend = pkgs.runCommand "faucet-frontend" {} ''
+    mkdir -p $out
+    sed -e 's|recaptchaSiteKey:.*|recaptchaSiteKey: "${recaptcha.siteKey}",|g' \
```
disassembler: I don't like that we're reading the recaptcha key from a private static area, but then injecting it plain text into the store. Essentially, if the server is hacked under any user, this makes the key visible. Couldn't we deploy this as a deploy key and have the faucet take a path to the file containing the key so it doesn't end up in the nix store? Since we own the faucet code, I don't see why we wouldn't want to go this route.

modules/cardano-faucet.nix
```diff
@@ -0,0 +1,55 @@
...
+    relayHost = "relays.cardano-testnet.iohkdev.io";
+    configKey = "testnet_full";
+    statsd.enable = true;
+    statsd.port = 8125; # dogstatsd will be listening on this port
+    inherit recaptcha;
+  };
+
+  networking.firewall.allowedTCPPorts = [ 80 ];
+
+  services.nginx = {
```
disassembler: lets set this up as SSL. `enableACME = true;` will get you a letsencrypt cert at a minimum

modules/cardano-faucet.nix
```diff
@@ -0,0 +1,55 @@
...
+  services.cardano-faucet = {
+    enable = true;
+    relayHost = "relays.cardano-testnet.iohkdev.io";
+    configKey = "testnet_full";
+    statsd.enable = true;
+    statsd.port = 8125; # dogstatsd will be listening on this port
+    inherit recaptcha;
+  };
+
+  networking.firewall.allowedTCPPorts = [ 80 ];
```
disassembler: I think we only want to expose 80 if we setup a forced ssl redirect (`forceSSL = true;`). 443 should be added here as well.

modules/cardano-faucet.nix
```diff
@@ -0,0 +1,55 @@
...
+
+  services.nginx = {
+    enable = true;
+    virtualHosts = let
+      vhostDomainName = if config.global.dnsDomainname != null
+        then config.global.dnsDomainname else "iohkdev.io";
+     in {
+      "cardano-faucet.${vhostDomainName}" = {
+        # TLS provided by cloudfront
```
disassembler: even if we have SSL terminating at cloudfront, it would still be good to have SSL from cloudfront to our server. https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-https-cloudfront-to-custom-origin.html

modules/cardano-faucet.nix
```diff
@@ -0,0 +1,55 @@
...
+        locations = {
+          "/" = {
+            root = faucetFrontend;
+            tryFiles = "$uri $uri/index.html @faucet";
```
disassembler: what about error pages? What do we want the user to see if the backend isn't available, or takes to long to respond, etc...

deployments/security-groups.nix
rvl: OK I will add 443 and leave 80 to allow redirecting http -> https.

lib/faucet.html
rvl: Will move it to `cardano-sl/faucet/frontend` and add a derivation there.

modules/cardano-faucet.nix
rvl: True. The recaptcha site key is public and injected into HTML. The recaptcha secret key is private and shouldn't be in the nix store. I already asked if faucet code could read the recaptcha secret key from a file but everything has to be asked for at least twice.

modules/cardano-faucet.nix
rvl: OK - will add both 80 and 443 and use forced ssl redirect.

modules/cardano-faucet.nix
rvl: Yes good point. Anyway I think cloudfront doesn't make sense for faucet but might for explorer (where this comment was copied from). So I will delete the comment and add enableACME.

modules/cardano-faucet.nix
rvl: The standard horrible error pages from NGINX should be OK.

This is because the faucet will normally be called by XHR through a page like this https://testnet.iohkdev.io/goguen/faucet/ . This frontend should show some "unavailable" message when it gets 5xx responses.

modules/cardano-faucet-module.nix
```diff
@@ -0,0 +1,268 @@
...
+    inherit (cfg) port;
+    payment-distribution = {
+      mean = cfg.paymentAmountMean;
+      scale = cfg.paymentAmountScale;
+    };
+    logging-config = cfg.loggingConfigFile;
+    source-wallet.generate-to = "${cfg.home}/generated-wallet.json";
+  } // (optionalAttrs (cfg.recaptcha != null) {
+    # fixme: recaptcha secret in file
+    recaptcha-secret = cfg.recaptcha.secretKey;
```
disassembler: good point on public vs secret recaptcha keys, but this is where the secret is getting dumped to the nix store.

rvl: Faucet has now been fixed to read the secret key from file so I can change this.

deployments/report-server-env-testnet.nix
rvl: Good point -- it is unneeded. I have removed this and killed the obsolete nixops resources.

rvl:
Thanks for the reviews. New commits have been pushed to this PR. I believe that I have addressed all review comments. The scripts are also shellcheck clean.

rvl:
Rebased on latest develop.

rvl:
Rebased onto master, resolved merge conflicts, removed change to `cardano-sl-src.json`, added CORS to faucet NGINX.

PR #362, opened 2018-06-20
[DEVOPS-819] Asset locking addresses for core nodes

disassembler:


modules/cardano.nix
```diff
@@ -101,7 +101,9 @@ wallet:
       publicIP = if options.networking.publicIPv4.isDefined then config.networking.publicIPv4 else null;
       privateIP = if options.networking.privateIPv4.isDefined then config.networking.privateIPv4 else "0.0.0.0";
       statsdServer = "127.0.0.1:8125";
-    };
+    }) // (optionalAttrs params.typeIsCore (let assetPath = ../static/asset-locked-addresses.nix; in {
```
rvl: Would it be better to fail hard if `static/asset-locked-addresses.nix` is not present and we are in the production environment? Trying to prevent the case of where the file is accidentally named wrong or something like that.

modules/cardano-service.nix
```diff
@@ -4,6 +4,7 @@ globals: params:
 { config, pkgs, lib, options, ... }:
 let
   cfg = config.services.cardano-node;
+  assetLockFile = pkgs.writeText "asset-lock-file" (intersperse "\n" cfg.assetLockAddresses);
```
rvl: Can the asset lock file be provided directly as a `.txt` file (`../static/asset-lock-file.txt`) rather than a nix expression?

disassembler:
@rvl switched to use a file instead of nix import. I'm hesitant to force usage of it yet, because I don't want to break reverse compatibility, especially if for some reason we need to deploy prior to us pushing 1.3.

rvl:
Is this good to merge?

PR #363, opened 2018-06-20
[DEVOPS-906] [DEVOPS-907] Update cardano-sl-src.json to release/1.3.0

rvl:
Using `cardano-sl.src` got broken after adding filtering of the sources.
See corresponding fix in input-output-hk/cardano-sl#3090.

PR #365, opened 2018-06-26
[DEVOPS-907] Fix cardano-service configuration

rvl:
Using `cardano-sl.src` got broken after adding filtering of the sources.
See corresponding fix in input-output-hk/cardano-sl#3090.


rvl:
I already made this PR in #363 ... oops.

PR #366, opened 2018-06-26
[DEVOPS-925] new machine: `hydra-mantis.iohk.io`

deepfire:
1. Factor the Hydra configuration into specific and common parts.
2. Add a second, non-distributed Hydra as `hydra-mantis.iohk.io`
3. Add the configuration bits specific to the existing Mantis Hydra deployment.


jobsets/secp256k1.nix
```diff
@@ -0,0 +1,24 @@
...
+
+stdenv.mkDerivation {
+  name = "secp256k1";
+
+  src = secp256k1Src;
+
+  buildInputs = [ autoconf automake libtool ];
+
+  configurePhase = ''
+    ./autogen.sh
```
cleverca22: adding `autoreconfHook` to the `nativeBuildInputs` will run autogen for you, and also bring autoconf and automake into scope for you
then the configure flags just need to be in `configureFlags`

jobsets/solidity.nix
```diff
@@ -0,0 +1,36 @@
...
+    makeWrapper
+  ];
+
+  patches = [ ./solidity.patch ];
+
+  configurePhase = ''
+    export NIX_CFLAGS_COMPILE="-Wno-error=maybe-uninitialized $NIX_CFLAGS_COMPILE"
+    mkdir build
+    cd build
+    cmake ..
```
cleverca22: the default `configurePhase` already does most of this when you add cmake to the inputs

modules/hydra-master-common.nix
```diff
@@ -0,0 +1,79 @@
...
+      patches = [
+        ./chomp.patch
+        ./hydra-nix-prefetch-git.patch
+        ./hydra-not-found.patch
+        ./hydra-github-pr-filter.patch
+      ];
+    });
+  };
+in {
+  environment.systemPackages = [ iohk-pkgs.iohk-ops ];
```
cleverca22: why is `iohk-ops` installed on hydra?

modules/hydra-master-mantis.nix
```diff
@@ -0,0 +1,87 @@
...
+      server_names_hash_bucket_size 64;
+
+      keepalive_timeout   70;
+      gzip            on;
+      gzip_min_length 1000;
+      gzip_proxied    expired no-cache no-store private auth;
+      gzip_types      text/plain application/xml application/javascript application/x-javascript text/javascript text/xml text/css;
+
+      server {
+        server_name _;
```
cleverca22: if this is switched to `nginx.virtualHosts.${hydraDnsName} = { ...` then you can just `enableACME = true; forceSSL = true;` and drop the entire `security.acme.certs` above and simplify a number of parts

deployments/infrastructure-env-production.nix
```diff
@@ -22,6 +22,23 @@ in {
     deployment.route53.hostName = "hydra.aws.iohkdev.io";
   };
 
+  mantis-hydra = { config, pkgs, resources, ... }: {
+    # Differences from the main Hydra:
+    # 1. builds are not sandboxed
+    # 2. hydra builds things without slaves
+    imports = [
+      ./../modules/papertrail.nix
```
cleverca22: redundant `./` is back

deployments/infrastructure-target-aws.nix
```diff
@@ -23,6 +23,25 @@ in rec {
       ];
     };
   };
+  mantis-hydra = { config, pkgs, resources, ... }: {
+    # See infrastructure-env-production.nix for description.
+    imports = [
+      ./../modules/amazon-base.nix
```
cleverca22: another `./` that can go

deployments/infrastructure-target-aws.nix
```diff
@@ -99,8 +118,9 @@ in rec {
       };
     };
     elasticIPs = {
-      hydra-ip    = { inherit region accessKeyId; };
-      cardanod-ip = { inherit region accessKeyId; };
+      hydra-ip        = { inherit region accessKeyId; };
+      mantis-hydra-ip = { inherit region accessKeyId; };
+      cardanod-ip     = { inherit region accessKeyId; };
     };
     datadogMonitors = (with (import ./../modules/datadog-monitors.nix); {
```
cleverca22: and here

jobsets/default-mantis.nix
```diff
@@ -1,47 +1,141 @@
...
+        })
+      ]
+      ++
+      (pkgs.lib.mapAttrsToList
+        (
+          num:
+          info: mkMantisJob {
+            name = "mantis-PR-${num}";
+            description = info.title;
+            mantisBranch = info.head.sha;
```
cleverca22: this means that a PR jobset cant auto-detect new pushes to the PR, it should be changed to use `pull/${num}/head` like cardano is using

jobsets/default-mantis.nix
```diff
@@ -1,47 +1,141 @@
...
+        (pkgs.lib.filterAttrs (num: info: info.base.ref == "phase/iele_testnet") mantisPrs)
+      )
+      ++
+      (pkgs.lib.mapAttrsToList
+        (
+          num:
+          info: mkMantisJob {
+            name = "solidity-service-PR-${num}";
+            description = info.title;
+            solidityServiceBranch = info.head.sha;
```
cleverca22: same as above

modules/hydra-master-main.nix
```diff
@@ -93,37 +68,28 @@ in {
...
       </githubstatus>
       <githubstatus>
-        jobs = serokell:cardano.*
+        jobs = iohk:cardano.*
         inputs = cardano
         excludeBuildFromContext = 1
       </githubstatus>
       <githubstatus>
-        jobs = serokell:daedalus.*:tests\..*
+        jobs = iohk:daedalus.*:tests\..*
```
cleverca22: once this change is deployed, hydra will cease to report status's to all PR's until the project name is fixed in the hydra state (the webui config)

disassembler: This is one I was hesitant about as well. Do we end up losing history if we do this? Do we need some sort of db migration path to make this change without losing history?

cleverca22: yeah, all history (including hydra metrics) would likely be lost, i dont believe hydra officially supports a rename, so you would have to create a new project, which starts fresh
declarative jobs would at least re-make every PR jobset, but all history will be missing

PR #370, opened 2018-07-05
[DEVOPS-779] simplify CI setup with buildkite hooks

disassembler:


modules/buildkite-agent.nix
```diff
@@ -14,8 +14,16 @@
     openssh.publicKeyPath  = "/run/keys/buildkite-ssh-public";
     tokenPath              = "/run/keys/buildkite-token";
     meta-data              = "system=x86_64-linux";
-    hooksPath              = "/var/lib/buildkite-agent/hooks";
     runtimePackages        = with pkgs; [ gnutar gzip bzip2 xz ];
+    hooks.pre-command = ''
+      NIX_PROFILE_BINPATH=/run/current-system/sw/bin
+      NIX_BUILD="$NIX_PROFILE_BINPATH/nix-build"
+      NIX_PATH="nixpkgs=$(''${NIX_BUILD} fetch-nixpkgs.nix -o nixpkgs)"
+      export NIX_REMOTE=daemon
```
rvl: This shouldn't be necessary I think.

modules/buildkite-agent.nix
```diff
@@ -14,8 +14,16 @@
     openssh.publicKeyPath  = "/run/keys/buildkite-ssh-public";
     tokenPath              = "/run/keys/buildkite-token";
     meta-data              = "system=x86_64-linux";
-    hooksPath              = "/var/lib/buildkite-agent/hooks";
     runtimePackages        = with pkgs; [ gnutar gzip bzip2 xz ];
+    hooks.pre-command = ''
+      NIX_PROFILE_BINPATH=/run/current-system/sw/bin
+      NIX_BUILD="$NIX_PROFILE_BINPATH/nix-build"
+      NIX_PATH="nixpkgs=$(''${NIX_BUILD} fetch-nixpkgs.nix -o nixpkgs)"
```
rvl: Pipeline being built should handle its own pinning

modules/buildkite-agent.nix
rvl: This shouldn't be necessary I think.

modules/buildkite-agent.nix
```diff
@@ -14,8 +14,16 @@
     openssh.publicKeyPath  = "/run/keys/buildkite-ssh-public";
     tokenPath              = "/run/keys/buildkite-token";
     meta-data              = "system=x86_64-linux";
-    hooksPath              = "/var/lib/buildkite-agent/hooks";
     runtimePackages        = with pkgs; [ gnutar gzip bzip2 xz ];
+    hooks.pre-command = ''
+      NIX_PROFILE_BINPATH=/run/current-system/sw/bin
```
rvl: What are NIX_PROFILE_BINPATH and NIX_BUILD for?
Could the PATH be set up with `/run/current-system/sw/bin`?

modules/buildkite-agent.nix
```diff
@@ -14,8 +14,16 @@
...
+      NIX_BUILD="$NIX_PROFILE_BINPATH/nix-build"
+      NIX_PATH="nixpkgs=$(''${NIX_BUILD} fetch-nixpkgs.nix -o nixpkgs)"
+      export NIX_REMOTE=daemon
+      export PATH="$NIX_PROFILE_BINPATH:$PATH"
+      export NIX_PATH
+      source /var/lib/buildkite-agent/hooks/pre-command
```
rvl: This seems unlikely to continue to work because it relies on state that's not set up by nixos configuration.

nix-darwin/modules/buildkite-agent.nix
```diff
@@ -27,6 +27,12 @@ in {
       else
         (>&2 echo "$creds doesn't exist. The build is going to fail.")
       fi
+      NIX_PROFILE_BINPATH=/nix/var/nix/profiles/default/bin
```
rvl: is `pre-command` the best hook for this? I thought it was there was a global environment hook.

modules/buildkite-agent.nix
rvl: Oh hang on, that file is set up by `deployment.keys`. Maybe change the filename.

nix-darwin/modules/buildkite-agent.nix
disassembler: There is an environment hook that's used for secrets. I guess it wouldn't hurt to move it there. Then any checkout/artifact parts of a job would also have access to nix.

PR #371, opened 2018-07-06
[DEVOPS-923] reduce mainnet eval memory usage by another 3.5gig

cleverca22:
https://gist.github.com/cleverca22/c99163e94b26d896f511d3540ce7303d

this switches the `shell.nix` env over to using the https://github.com/nixos/nix/tree/memoise branch of nix, and then uses it to deduplicate the mainnet deployment some
this first change shaves 3.5gig off the memory usage, bringing it from 34.8gig to 31.3gig

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -25,6 +25,19 @@ let
     in (import "${nixopsUnstable}/release.nix" {
          nixpkgs = localLib.fetchNixPkgs;
         }).build.${system};
+  nix-src = pkgs.fetchFromGitHub {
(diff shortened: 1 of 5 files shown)


disassembler:
I approve of this, but wonder if the 3.5 GB saved of RAM is going to make that much of a difference in the long run.

cleverca22:
@disassembler it also shaves 107 seconds (31% of the time) off the eval of the deployment

PR #374, opened 2018-07-10
DEVOPS-193:  cardano-node-commit-monitor

deepfire:
Not exactly what was envisioned, but in the same ballpark.

### DIFF
--- a/iohk/NixOps.hs
+++ b/iohk/NixOps.hs
@@ -851,8 +851,9 @@ scpFromNode o c (fromNodeName -> node) from to = do
     ExitFailure code -> TIO.putStrLn $ "scp from " <> node <> " failed with " <> showT code
 
 triggerCommitMonitor :: Options -> NixopsConfig -> NodeName -> IO ()
-triggerCommitMonitor o c m = do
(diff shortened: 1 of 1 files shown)


iohk/NixOps.hs
```diff
@@ -844,6 +855,10 @@ scpFromNode o c (fromNodeName -> node) from to = do
     ExitSuccess -> return ()
     ExitFailure code -> TIO.putStrLn $ "scp from " <> node <> " failed with " <> showT code
 
+triggerCommitMonitor :: Options -> NixopsConfig -> NodeName -> IO ()
+triggerCommitMonitor o c m = do
+  ssh' o c "bash" ["-c", "strings $(pgrep -fal cardano-node-simple | cut -d' ' -f2) 2>/dev/null | egrep '^[0-9a-f]{40,40}$' | xargs logger -t cardano-node-commit-monitor"] m empty
```
disassembler: If we do decide to go forward with this PR, lets change logger to: `systemd-cat -t cardano-node-commit-fetcher -p info`

PR #380, opened 2018-07-26
[DEVOPS-859] nix-darwin improvements

disassembler:
I'm using this locally on my home network to manage my macs and I've made some improvements I thought we might want in iohk-ops.

### DIFF
--- a/nix-darwin/NukeNix.hs
+++ b/nix-darwin/NukeNix.hs
@@ -1,18 +1,22 @@
 #! /usr/bin/env nix-shell
 #! nix-shell -i runhaskell
-{-# LANGUAGE OverloadedStrings, LambdaCase #-}
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 -- Use this script to reset all/most of the nix and nix-darwin
 -- changes. This is just for debugging the deployment script.
 -- Run with sudo.
 
-import Prelude hiding (FilePath)
-import Turtle
-import Data.Text (Text)
-import Control.Monad (forM_)
+module NukeNix where
 
-main :: IO ()
-main = sh $ do
+import           Control.Monad      (forM_)
+import           Data.Text          (Text)
+import           Prelude            hiding (FilePath)
+import           System.Environment
+import           Turtle
+
+nuke :: IO ()
+nuke = sh $ do
   sh $ do
     let pat = star dot *> tab *> star dot *> tab *> begins "org.nixos"
     daemon <- inproc "launchctl" ["list"] empty & grep pat & sed pat
@@ -28,6 +32,7 @@ main = sh $ do
                     , ".nixpkgs"
                     , ".config/nixpkgs"
                     ]
+  homeDir <- liftIO $ getEnv "HOME"
   let dead = [ "/etc/nix"
              , "/etc/bashrc.backup-before-nix-darwin"
              , "/nix"
@@ -41,7 +46,8 @@ main = sh $ do
                , "org.nixos.buildkite-agent.plist"
                , "org.nixos.nix-gc.plist"
                ]) ++
-             (under "/Users/admin" userConfigs) ++
+             (under (fromString homeDir) userConfigs) ++
+             (under "/var/root" userConfigs) ++
              (under "/root" userConfigs)
 
   void $ proc "rm" ("-rf":files dead) empty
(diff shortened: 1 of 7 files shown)


nix-darwin/NukeNix.hs
```diff
@@ -28,6 +32,7 @@ main = sh $ do
                     , ".nixpkgs"
                     , ".config/nixpkgs"
                     ]
+  homeDir <- liftIO $ getEnv "HOME"
```
rvl: Can use `Turtle.home`

nix-darwin/PrepareDarwin.hs
```diff
@@ -78,7 +87,11 @@ setupSSLCert = unlessM (testfile cert) $ mapM_ sudo setup
 prepareConfigs :: Shell ()
 prepareConfigs = do
   -- prepare configs for nix darwin
-  mapM_ moveAway ["/etc/bashrc", "/etc/nix/nix.conf"]
+  user <- liftIO $ getEnv "USER"
```
rvl:
```
need "USER" >>= \case
    Just user -> do { ... }
    Nothing -> echo "USER not set"
```

nix-darwin/PrepareDarwin.hs
```diff
@@ -101,9 +124,10 @@ chopProfile p = do
   sudo ["cp", tt temp, tt p]
 
 -- | This is needed so that nix-copy-closure to this host will work
-nixCopyClosureHack :: Shell ()
-nixCopyClosureHack = unlessM (testpath "/usr/bin/nix-store") $
-  sudo ["ln", "-sf", "/nix/var/nix/profiles/default/bin/nix-store", "/usr/bin"]
+setupUserBashrc :: IO ()
+setupUserBashrc = do
+  homeDir <- getEnv "HOME"
```
rvl:
```
bashRC <- (</> ".bashrc") <$> home
```

Good fix -- better than symlinking `/usr/bin/nix-store` :-)

nix-darwin/PrepareDarwin.hs
```diff
@@ -135,6 +170,8 @@ setupBuildkiteDir bk = do
     printf ("Placeholder token file: " % fp % "\n") token
     liftIO $ writeTextFile token ""
 
+
+-- | Create a directory for secret files which shouldn't go in nix store.
```
rvl: This comment is out of place.

nix-darwin/RewriteLibs.hs
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env nix-shell
```
rvl: This module has a shebang but no main function. Maybe set this up as a `rewrite-libs` program with main and replace call to `patch-prepare` with `rewrite-libs --out-dir $out/bin $out/bin/*`

nix-darwin/everything.hs
```diff
@@ -0,0 +1,16 @@
+module Main where
+
+import           Data.Monoid
+import           NukeNix
+import           PrepareDarwin
+import           System.Environment
+
+main :: IO ()
+main = do
```
rvl: Might be more direct to use LambdaCase and write
```
main = progname >>= \case
  "prepare" -> prepare
  "nuke-nix" -> nuke
  _ -> echo "wtf"
```

nix-darwin/PrepareDarwin.hs
```diff
@@ -76,10 +85,16 @@ setupSSLCert = unlessM (testfile cert) $ mapM_ sudo setup
...
 prepareConfigs :: Shell ()
-prepareConfigs = do
+prepareConfigs =
   -- prepare configs for nix darwin
-  mapM_ moveAway ["/etc/bashrc", "/etc/nix/nix.conf"]
-  chopProfile "/etc/profile"
+  need "USER" >>= \case
+    Just user -> do
+      mapM_ moveAway ["/etc/nix/nix.conf"]
+      let contents = fromString $ "trusted-users = " <> T.unpack user
```
cleverca22: if you get rid of both the `fromString` and the `T.unpack` it should remain as a `Text` and not need any conversion

nix-darwin/PrepareDarwin.hs
```diff
@@ -135,6 +171,7 @@ setupBuildkiteDir bk = do
     printf ("Placeholder token file: " % fp % "\n") token
     liftIO $ writeTextFile token ""
 
+
```
cleverca22: extra newline added

PR #384, opened 2018-07-30
ci | bench:  don't test the bench env by default:  too slow to gen 50k keys

deepfire:
@365andreas, that 50k key generation in the benchmark test is slow -- what should we do about it?

This disables bench environment testing by default.


deepfire:
@365andreas, the problem is with `cardano-keygen`, which only cares about the particular `bench` key under `lib/configuration.yaml`.

I.e. the change you suggest will have no effect.

disassembler:
went with https://github.com/input-output-hk/iohk-ops/pull/385

PR #388, opened 2018-08-01
[DEVOPS-986] disable auto-restart on mainnet r-a-1

cleverca22:


deployments/cardano-nodes.nix
```diff
@@ -1,7 +1,9 @@
...
+{ environment, globals, ... }: with (import ./../lib.nix);
 let nodeMap = globals.nodeMap; in
 
-flip mapAttrs nodeMap
+(flip mapAttrs nodeMap
 (name: import ./../modules/cardano.nix
        globals
-       [])
+       [])) // {
+  require = if (environment == "production") && (nodeMap ? r-a-1) then [ ./mainnet-noauto-restart.nix ] else [];
```
jmitchell: I realize `nodeMap ? r-a-1` was added to fix a broken test, but as a result `r-a-1` is duplicated.

We may also decide to adjust the set of nodes where cardano-restart is disabled, so it would be nice to support an explicit list, e.g. ``rollingRestartBlacklist = [r-a-1];`, and remove any other mentions of specific hosts for this feature.

PR #390, opened 2018-08-02
[DEVOPS-987] Replacing ci shellscript with turtle

disassembler:


.buildkite/hooks/pre-command
```diff
@@ -0,0 +1,5 @@
+=======
```
cleverca22: line 1 still needs to be fixed

scripts/ci.hs
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env nix-shell
+#!nix-shell ../shell.nix -A ci --run "runhaskell -iiohk scripts/ci.hs"
```
cleverca22: once testing is done, it should probably be switched to a nix derivation that is pre-built by hydra

scripts/ci.hs
```diff
@@ -0,0 +1,120 @@
...
+   , sStaging         :: Bool
+   , sProduction      :: Bool
+   , sInfra           :: Bool
+   , sExplorer        :: Bool
+   , sReportServer    :: Bool
+   }
+settingsParser :: Parser Settings
+settingsParser =
+    Settings <$> (optPath "iohk-ops" 'o' "Path to `iohk-ops` binary" <|> pure FP.empty)
+             <*> switch  "cleanup-configs" 'c' "Clean up configs"
```
cleverca22: has the bug with switch been fixed yet?

PR #393, opened 2018-08-06
[DEVOPS-788] [DEVOPS-868] Update proposal Linux support and signing fix

rvl:
* [DEVOPS-788](https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-788)
   - Adds support for proposing linux installers with `io update-proposal submit`.
   - Add a basic test case for checking the `cardano-auxx propose-update` command line.
* [DEVOPS-868](https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-868)
   - Pass tty to `gpg2` subprocess so that password entry works.
   - Make `io update-proposal find-installers -u user-id` optional so that the default signing key from the keyring can be used.
   - Move sha256 and blake2b hashing of installers earlier to the `find-installers` step.

### DIFF
--- a/iohk/UpdateProposal.hs
+++ b/iohk/UpdateProposal.hs
@@ -42,7 +42,7 @@ import Types ( NixopsDepl(..), Environment(..), Arch(..)
              , archMapToList, archMapFromList, lookupArch
              , idArchMap, archMapEach)
 import UpdateLogic ( InstallersResults(..), CIResult(..)
-                   , realFindInstallers, githubWikiRecord
+                   , getInstallersResults, githubWikiRecord
                    , runAWS', uploadHashedInstaller, updateVersionJson
                    , uploadSignature )
 import RunCardano
@@ -441,7 +441,7 @@ updateProposalFindInstallers opts env = do
   echo "*** Finding installers"
   let rev = unGitRevision . cfgDaedalusRevision $ params
       destDir = Just (installersDir opts)
-  res <- liftIO $ realFindInstallers (configurationKeys env) (installerForEnv env) rev destDir
+  res <- liftIO $ getInstallersResults (configurationKeys env) (installerForEnv env) rev destDir
 
   echo "*** Hashing installers with sha256sum"
   sha256 <- getHashes sha256sum res
--- a/iohk/UpdateProposal.hs
+++ b/iohk/UpdateProposal.hs
@@ -69,6 +69,7 @@ data UpdateProposalStep
   | UpdateProposalSubmit
     { updateProposalRelayIP           :: Text
     , updateProposalDryRun            :: Bool
+    , updateProposalWithSystems       :: ArchMap Bool
     }
   deriving Show
 
@@ -93,7 +94,7 @@ parseUpdateProposalCommand = subparser $
        (info ((UpdateProposalCommand <$> date <*> pure UpdateProposalGenerate) <**> helper)
          (progDesc "Create a voting transaction") ) )
   <> ( command "submit"
-       (info (UpdateProposalCommand <$> date <*> (UpdateProposalSubmit <$> relayIP <*> dryRun))
+       (info (UpdateProposalCommand <$> date <*> (UpdateProposalSubmit <$> relayIP <*> dryRun <*> withSystems))
          (progDesc "Send update proposal transaction to the network") ) )
   where
     date :: Parser (Maybe Text)
@@ -132,6 +133,11 @@ parseUpdateProposalCommand = subparser $
     dryRun :: Parser Bool
     dryRun = switch ( long "dry-run" <> short 'n' <> help "Don't actually do anything" )
 
+    withSystems :: Parser (ArchMap Bool)
+    withSystems = mkArchMap <$> withLinux <*> pure True <*> pure True
+      where withLinux = switch ( long "with-linux" <>
+                                 help "Also propose a Linux installer" )
+
 updateProposal :: Options -> NixopsConfig -> UpdateProposalCommand -> IO ()
 updateProposal o cfg UpdateProposalCommand{..} = do
   configKey <- maybe (fail "configurationKey not found") pure (nixopsConfigurationKey cfg)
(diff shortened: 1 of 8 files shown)


iohk/UpdateProposal.hs
```diff
@@ -270,33 +276,47 @@ instance Checkable UpdateProposalConfig1 where
...
+  checkConfig (UpdateProposalConfig2 p r b s) = checkConfig p <|> check r
+                                                <|> checkInstallerHashes b
+                                                <|> checkInstallerHashes s
     where
       check InstallersResults{..}
         | grApplicationVersion <= 0 = Just "Application version must be set"
         | T.null grCardanoCommit = Just "Missing cardano commit"
         | missingVersion Mac64 ciResults = Just "macOS version is missing"
         | missingVersion Win64 ciResults = Just "Windows version is missing"
+        | missingVersion Linux64 ciResults = Just "Linux version is missing"
```
disassembler: won't this require a linux version to be passed (instead of it being optional)?

iohk/UpdateProposal.hs
```diff
@@ -454,45 +485,42 @@ updateProposalUploadS3 :: CommandOptions -> Shell ()
...
       printf ("***   "%s%"  "%fp%"\n") hash (ciResultLocalPath ci)
       uploadHashedInstaller bucket (ciResultLocalPath ci) (globalResult res) hash
 
 -- | Apply a hashing command to all the installer files.
 getHashes :: (FilePath -> Shell Text) -> InstallersResults -> Shell InstallerHashes
 getHashes getHash res = do
-  cfgInstallerDarwin <- check Mac64 =<< needResult Mac64 res resultHash
-  cfgInstallerWindows <- check Win64 =<< needResult Win64 res resultHash
-  pure $ InstallerHashes{..}
+  linux64 <- check Win64 =<< needResult Linux64 res resultHash
```
disassembler: s/Win64/Linux64 on this line

iohk/UpdateProposal.hs
rvl: I was thinking that because Linux installers have been reliably built in buildkite for some time, and that Linux will be fully supported in 1.4.0, there is no need to make it optional.

Does that sound alright, or should I add a flag or something?

iohk/UpdateProposal.hs
rvl: Ah yes thanks. I didn't have the energy to reformulate `needResult` but maybe I should to avoid this duplication.

cleverca22:
https://gist.github.com/cleverca22/d903059a0a4827ee14676f8d5bc67908 
this is an example i threw together on how we can deal with taking a list of arch+installer pairs, and generate an update-proposal command, with no chance of getting the arch+installer pairs cross-wired

rvl:
Yes thanks, I had that example in mind when writing this.
I think `archUpdateBin` should be equivalent to that.

rvl:
Rebased on latest master to resolve merge conflicts.

rvl:
Rebased again, and made proposing of Linux installer optional (using `io update-proposal submit --with-linux`).

iohk/UpdateProposal.hs
```diff
@@ -619,6 +655,11 @@ copyInstallerFiles opts res InstallerHashes{..} = do
     file :: FilePath -> Shell Text
     file dst = snd <$> procStrict "file" [tt dst] empty
 
+    installerMagic :: Arch -> Text
```
deepfire: Should this be an `ArchMap`?

iohk/UpdateProposal.hs
```diff
@@ -581,15 +609,23 @@ updateProposalSetVersionJSON cfg opts@CommandOptions{..} = do
...
-  copyInstallerFiles opts (cfgInstallersResults cfgUpdateProposal2) cfgInstallerHashes
+  copyInstallerFiles opts cfgInstallersResults cfgInstallerHashes
   echo "*** Generating keys and database."
   addrs <- doGenerate opts params
   storeParams opts (UpdateProposalConfig4 params addrs)
   echo "*** Finished generate step. Next step is to submit."
   echo "*** Carefully check the parameters yaml file."
 
+needResults :: MonadIO io => InstallersResults -> (Arch -> CIResult -> io a) -> io (ArchMap a)
+needResults rs action = do
```
deepfire: If we make `ArchMap` an `Applicative`, this could be simplified, I believe.

iohk/UpdateProposal.hs
```diff
@@ -509,55 +543,49 @@ uploadResultSignature bucket res = liftIO $ maybeReadFile sigFile >>= \case
...
+        attrs = [ (keyPrefix arch <> k, String v) | (k, v) <-
                     [ (""         , dviVersion)
                     , ("URL"      , dviURL)
                     , ("Hash"     , dviHash)
                     , ("SHA256"   , dviSHA256)
                     , ("Signature", fromMaybe "" dviSignature)
                     ] ]
+    keyPrefix Mac64   = "macos"
+    keyPrefix Win64   = "win64"
+    keyPrefix Linux64 = "linux"
```
deepfire: Also an `ArchMap`?

iohk/UpdateProposal.hs
```diff
@@ -270,33 +282,47 @@ instance Checkable UpdateProposalConfig1 where
...
         | T.null grCardanoCommit = Just "Missing cardano commit"
         | missingVersion Mac64 ciResults = Just "macOS version is missing"
         | missingVersion Win64 ciResults = Just "Windows version is missing"
+        | missingVersion Linux64 ciResults = Just "Linux version is missing"
+        | length (nub as) /= length as = Just "Multiple installers found for an arch"
         | otherwise = Nothing
         where
           GlobalResults{..} = globalResult
           missingVersion arch = not . any ((== arch) . ciResultArch)
+          as = map ciResultArch ciResults
```
deepfire: I think the above code could've been simplified if `ciResults` was an `ArchMap`, making the structural checks unnecessary.

deepfire:
I don't want to throw a spanner in the works, but I think the PR could have simplified some of the code if the `ArchMap` abstraction it introduces (a very nice idea!) was used in a more principled way.

That way, we could get rid of some structural checks and explicit architecture mentions sprinkled all around the code.

rvl:
@deepfire OK good idea.

I have made `ArchMap` an `Applicative` and fixed the arch-specific checks by building `ArchMap CIResult`.

There is no need to change occurrences of `Arch -> a` to `ArchMap a`. They are isomorphic.

PR #397, opened 2018-08-10
[DEVOPS-407]  update-proposal | cdn:  use CDN urls for the website

deepfire:


iohk/common/Utils.hs
```diff
@@ -112,3 +114,8 @@ jsonLowerStrip n = AE.genericToJSON $ AE.defaultOptions { AE.fieldLabelModifier
 s3Link :: Region -> BucketName -> ObjectKey -> Text
 s3Link region (BucketName bucket) (ObjectKey key) =
   mconcat [ "https://s3-", AWS.toText region, ".amazonaws.com/" , bucket, "/", key ]
+
+cdnLink :: HasCallStack => Environment -> ObjectKey -> Text
```
cleverca22: what about making this an entry in the `config.yaml` so development clusters support updating
this code makes it harder to test the whole update system in a disposable cluster

deepfire: To be fair, this code doesn't really change the status quo -- we had `s3Link` to the same effect previously.

deepfire: I agree that your proposal is desirable, though!

deepfire: Done..

PR #400, opened 2018-08-15
[DEVOPS-868] Update proposal automation fixes

rvl:
https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-868

* Makes the _date_ argument to `io update-proposal` commands compulsory to avoid problems running the procedure around UTC midnight.
* Roll the `generate` step into `submit`.
* Fix pattern matching of the update proposal ID which got broken by cardano-sl 1.3.0.
* Require a single CI build per commit. If there is more than one build found (e.g. due to multiple pull requests for the same revision), then the user has to provide a build number.

### DIFF
--- a/iohk/common/Arch.hs
+++ b/iohk/common/Arch.hs
@@ -37,7 +37,6 @@ formatArch Mac64 = "macOS"
 formatArch Win64 = "Windows"
 
 -- | A map of values indexed by Arch.
--- Maybe "type ArchMap a = Arch -> a" would be better but this works.
 data ArchMap a = ArchMap { linux64 :: !a, mac64 :: !a, win64 :: !a }
   deriving (Show, Read, Eq, Generic, Functor, Foldable, Traversable)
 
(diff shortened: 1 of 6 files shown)


iohk/common/Arch.hs
```diff
@@ -0,0 +1,99 @@
...
+
+type ApplicationVersionKey = Arch -> Text
+
+formatArch :: Arch -> Text
+formatArch Linux64 = "Linux"
+formatArch Mac64 = "macOS"
+formatArch Win64 = "Windows"
+
+-- | A map of values indexed by Arch.
+-- Maybe "type ArchMap a = Arch -> a" would be better but this works.
```
angerman: Could this state *why* `Arch -> a` would be better, or maybe just omit the comment?

My reasoning is that the next person reading this code will wonder why the original
author left that comment.

rvl: OK removed the confusing comment.

An assoc list or Map could have missing keys but ArchMap can't.

Really it doesn't matter too much about the implementation. This one might be more boilerplate than necessary but it's fairly straightforward.

PR #407, opened 2018-08-22
cardano-service:  fix

deepfire:


rvl:
How do the start and stop commands work? I don't get why `Start` needed to be added.
It would also help to have some kind of description (in commit or PR) of why this change is needed.

PR #408, opened 2018-08-29
[DEVOPS-1012] enforce a 2 hour timeout on linux build slaves

cleverca22:


modules/hydra-slave.nix
```diff
@@ -14,10 +13,13 @@ in
 
   nix = {
     binaryCaches = mkForce [ "https://cache.nixos.org" ];
+    extraOptions = ''
+      # max of 2 hours for any given derivation on linux
+      # note darwin is slower, and should have a higher timeout, maybe 4h?
```
rvl: For a single derivation, 2 hours should be plenty, shouldn't it? Are there any builds on macOS which are exceeding that?

cleverca22:
upstream issue also filed: https://github.com/NixOS/hydra/issues/591

PR #412, opened 2018-09-05
[DEVOPS-1039] add influxdb/grafana and clean up nginx config for hydra

disassembler:
Adds grafana and influxdb to hydra master. Cleans up nginx to make it easier to work with in the future. grafana should have google oauth support, although untested as of yet.

### DIFF
--- a/modules/hydra-master.nix
+++ b/modules/hydra-master.nix
@@ -124,6 +124,20 @@ in {
     package = pkgs.postgresql96;
     dataDir = "/var/db/postgresql-${config.services.postgresql.package.psqlSchema}";
   };
+  services.influxdb = {
+    enable = true;
+  };
+  services.grafana = {
+    enable = true;
+    users.allowSignUp = true;
+    domain = "hydra.iohk.io";
+    rootUrl = "%(protocol)s://%(domain)s/grafana/";
+    extraOptions = {
+      AUTH_GOOGLE_ENABLED = "true";
+      AUTH_GOOGLE_CLIENT_ID = "778964826061-5v0m922g1qcbc1mdtpaf8ffevlso2v7p.apps.googleusercontent.com";
+      AUTH_GOOGLE_CLIENT_SECRET = builtins.readFile ../static/google_oauth_hydra_grafana.secret;
+    };
+  };
 
   systemd.services.hydra-evaluator.path = [ pkgs.gawk ];
   systemd.services.hydra-manual-setup = {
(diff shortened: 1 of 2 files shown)


modules/hydra-master.nix
```diff
@@ -153,57 +167,37 @@ in {
...
+        '';
+        locations."/graph/".extraConfig = ''
+          proxy_pass http://127.0.0.1:8081/;
+        '';
+        locations."/grafana/".extraConfig = ''
+          proxy_pass http://localhost:3000/;
+        '';
+      };
+    };
+    commonHttpConfig = ''
```
rvl: We could leverage `services.nginx.recommendedGzipSettings` and `recommendedOptimisation` although these are slightly different settings which would need to be tested, so no problem to keep it as is.

modules/hydra-master.nix
```diff
@@ -153,57 +167,37 @@ in {
...
-        }
-        location ~ /(nix-cache-info|.*\.narinfo|nar/*) {
+        '';
+        locations."~ /(nix-cache-info|.*\\.narinfo|nar/*)".extraConfig = ''
           return 301 https://iohk-nix-cache.s3-eu-central-1.amazonaws.com$request_uri;
-        }
-      }
+        '';
+        locations."/graph/".extraConfig = ''
+          proxy_pass http://127.0.0.1:8081/;
```
rvl: Where is this set up?

modules/hydra-master.nix
```diff
@@ -124,6 +124,20 @@ in {
...
+    enable = true;
+  };
+  services.grafana = {
+    enable = true;
+    users.allowSignUp = true;
+    domain = "hydra.iohk.io";
+    rootUrl = "%(protocol)s://%(domain)s/grafana/";
+    extraOptions = {
+      AUTH_GOOGLE_ENABLED = "true";
+      AUTH_GOOGLE_CLIENT_ID = "778964826061-5v0m922g1qcbc1mdtpaf8ffevlso2v7p.apps.googleusercontent.com";
```
rvl: Are these credentials attached to a shared google account set up by IT?

disassembler: it's currently on my google account. Figure we can test and make sure it works before involving Jeremy to create an oauth client at the org level and then just PR the change of the client id and update the secret.

modules/hydra-master.nix
disassembler: This was the stuff @cleverca22 was working on that was deployed to hydra but not PR'd already. I included it in this PR so I don't revert the changes since he's using that endpoint for developing the frontend svg generator locally with that data.

modules/hydra-master.nix
disassembler: I thought about that, but as you said, they vary slightly and would need much more testing. Once this refactor is in, it's much easier to update going forward though.

PR #418, opened 2018-09-10
[DEVOPS-1032] hydra: Build cardano-sl PRs with optimization disabled

rvl:
See also: input-output-hk/cardano-sl#3544.


rvl:
Merged, thanks.
It will apply the extra argument to the nix-build of the release expression, but this will be silently ignored until the corresponding change appears in cardano-sl.

cleverca22:
this has entirely broken all declarative jobsets, because `fasterBuild` is a boolean, when it must be a set describing the input

rvl:
Oops! Thanks

PR #425, opened 2018-09-13
[DEVOPS-1039] Github Webhook Utility for measuring CI times

disassembler:


github-webhook-util/Main.hs
```diff
@@ -0,0 +1,274 @@
...
+        putStrLn "Importing commit"
+        mapM_ processStatuses $ HM.toList $ ghStatusMap acc
+
+main :: IO ()
+main = do
+  self <- getProgName
+  let
+      go :: String -> IO ()
+      go "servant" = servantMain
+      go "manual"  = manualInfluxImport
```
cleverca22: if you adjust these names, and then make a symlink in the `$out/bin` then you can access both mains

github-webhook-util/Main.hs
```diff
@@ -0,0 +1,274 @@
...
+  , avFirstStatus :: Maybe UTCTime
+  } deriving Show
+
+data Accumulator = Accumulator {
+    ghStatusMap :: HM.HashMap (GH.Name GH.Commit) AccumulatorValue
+  } deriving Show
+
+getAuth :: IO (Maybe GH.Auth)
+getAuth = do
+    token <- lookupEnv "GITHUB_TOKEN"
```
jmitchell: Let's document the least privileges required for this OAuth token and ensure the one we use is configured that way.

PR #431, opened 2018-09-21
[CBR-275] Adapt to new logging

365andreas:
Adapt code and documentation to the new logging and to changes made in master.


CodiePP:
added the correct selection of log files to be transferred from the node.
rebased.

PR #440, opened 2018-10-01
[DEVOPS-1066] Add bash completion to `io` command

rvl:
Dug up my old branch with bash completion enabled because it can be part of some general work to improve the iohk-ops CLI.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -5,12 +5,14 @@
 let
   iohkpkgs = import ./default.nix {};
   iohk-ops = iohkpkgs.iohk-ops;
-  justIo = pkgs.runCommand "shell" {
+  justIo = pkgs.mkShell {
+    name = "io";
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -21,11 +23,22 @@ let
...
+    name = "io-with-auxx";
     buildInputs = [
       iohk-ops
       iohkpkgs.cardano-sl-auxx
       iohkpkgs.cardano-sl-tools
     ];
-  } "echo use nix-shell";
+    shellHook = ''
+      ${ioShellHook}
+      for prog in cardano-keygen cardano-x509-certificates cardano-auxx; do
```
angerman: can't we pull them out of the buildInputs somehow?

rvl: We could have, except that I added a program to cardano-sl-tools which doesn't use optparse-applicative...

PR #445, opened 2018-10-03
[DEVOPS-859] fixing nixStable from correct pkgs attribute

disassembler:


disassembler:
The added `nixStable` I don't think is needed anymore @rvl but please confirm. If it is needed, I can switch it to `pkgs_mac.nixStable` instead of removing it, but I'm not sure unless you're running the shell on OSX it's useful at all to shell into this expression anymore.

nix-darwin/default.nix
```diff
@@ -8,7 +8,7 @@ in rec {
     ghc = pkgs_mac.haskellPackages.ghcWithPackages (ps: [ ps.turtle ps.universum ps.megaparsec ]);
   in pkgs_mac.stdenv.mkDerivation {
     name = "deploy-nix-darwin";
-    buildInputs = [ ghc pkgs.nixStable ];
+    buildInputs = [ ghc ];
     shellHook = "eval $(egrep ^export ${ghc}/bin/ghc)";
```
cleverca22: and if its not useful as a shell expression, maybe also remove the `shellHook`?

disassembler: @rvl said it's still useful, so left it in.

rvl:
- `nixStable` is needed in the shell for the reason stated in baa2439bea49cabb2c36c0c61577c1c51348bdf2.
- The `shellHook` is needed so that the Haskell dependencies are available to ghc/ghci when running a nix-shell.

PR #449, opened 2018-10-04
[DEVOPS-904] add release notes to the version.json

cleverca22:


iohk/UpdateProposal.hs
```diff
@@ -113,14 +115,19 @@ parseUpdateProposalCommand = subparser $
...
+    revisionP = option (eitherReader (gitRevision . T.pack)) (long "revision" <> short 'r' <> metavar "SHA1" <> help "Daedalus revision to fetch")
+
+    blockVersionP :: Parser Text
+    blockVersionP = fmap T.pack $ strOption (long "block-version" <> short 'B' <> metavar "VERSION" <> help "Last known block version. Check the wiki for more info.")
+
+    voterIndexP :: Parser Int
+    voterIndexP = option auto (long "voter-index" <> short 'V' <> metavar "INTEGER" <> help "A number representing you, the vote proposer. Check the wiki for more info.")
+
+    releaseNotesP :: Parser (Maybe Text)
+    releaseNotesP = (Just . T.pack <$> strOption ( long "release-notes" <> metavar "RELEASE_NOTES" <> (help "Path to release notes (html)") <> (completer (bashCompleter "file") )))
```
rvl: Better use the `FilePath` for all paths and `FP.fromText <$> option auto ...` to parse the option.

iohk/UpdateProposal.hs
```diff
@@ -414,6 +426,7 @@ versionFile opts = cmdWorkPath opts </> "daedalus-latest-version.json"
 -- | Step 1. Init a new work directory
 updateProposalInit :: FilePath -> UpdateID -> UpdateProposalConfig1 -> Shell ()
 updateProposalInit top uid@(UpdateID _ date) cfg = do
+  liftIO $ print cfg
```
rvl: don't need this debugging statement

iohk/UpdateProposal.hs
```diff
@@ -423,7 +436,12 @@ updateProposalInit top uid@(UpdateID _ date) cfg = do
     True -> die "Config file already exists, stopping."
     False -> pure ()
   printf ("*** Creating blank template for date "%s%"\n") date
-  liftIO $ encodeFile (encodeString yaml) cfg
+  cfg' <- case (cfgReleaseNotes cfg) of
```
rvl: could this be separated into a new function?

iohk/UpdateProposal.hs
cleverca22: the problem, is that i need the parser to return a `FilePath`, but then i want to store the `Text` within that file in the data record
and then i need 2 data records, for the parser, and the result after applying IO

rvl: OK fair enough

rvl:
It would be better if the code to read release notes and update the config were in a separate function.

rvl:
I'm merging this now so that we can try it for the staging update proposal but the [docs](https://github.com/input-output-hk/iohk-ops/blob/master/docs/Daedalus-Installer-Update-Procedure.md) also need to be updated.

@cleverca22 Please update the docs in a separate PR

bors r+

PR #451, opened 2018-10-05
[DEVOPS-1032] jobsets: extra jobset to build cardano-sl develop branch without optimization

rvl:
The `no-opt-cardano-sl` jobset builds without optimization, so that PR builds based on `develop` will be able to use prebuilt packages.

The jobset name starts with `no-opt` so that it's not picked up by the Hydra github status plugin.

### Testing
```
nix-build jobsets/default.nix -A jobsets -o spec.json
jq '.["cardano-sl"]' < spec.json
jq '.["no-opt-cardano-sl"]' < spec.json
jq '.["cardano-sl-pr-42"]' < spec.json
```


disassembler:
bors test

PR #459, opened 2018-10-23
jobsets: fetch submodules with cardano-sl

rvl:


rvl:
Wrong option, it needs a patched Hydra.

PR #463, opened 2018-11-02
[DEVOPS-1103] log classifier service

disassembler:


deployments/infrastructure-env-production.nix
```diff
@@ -12,15 +10,14 @@ let region = "eu-central-1";
...
-      deployment.ec2.elasticIPv4 = resources.elasticIPs."${hostname}-ip";
-
-      deployment.route53.accessKeyId = route53accessKeyId;
-      deployment.route53.hostName = "${hostname}.aws.iohkdev.io";
     };
 in {
   network.description = "IOHK infrastructure production";
 
+  defaults = {
+    services.dd-agent.tags = ["env:production" ];
```
deepfire: I think there's duplication with line 12 here.

deepfire: Are we sure this doesn't break the `depl` tag?

deployments/infrastructure-env-staging.nix
```diff
@@ -1,28 +1,24 @@
...
 
 with (import ./../lib.nix);
-let region = "eu-central-1";
-    accessKeyId = IOHKaccessKeyId;
-in {
+{
   network.description = "IOHK infrastructure staging";
 
+  defaults = {
+    services.dd-agent.tags = ["env:staging" ];
```
deepfire: Again, breaks `depl` tag.

PR #469, opened 2018-11-08
cardano-sl:  bump for rc2

deepfire:


deepfire:
@rvl, it corresponds to the RC2 email, which is the intent.

PR #471, opened 2018-11-08
hydra-main:  use same commit as hydra-mantis

deepfire:
Mantis Hydra is submodule-capable.  Let's make our main Hydra as good.


jmitchell:
@deepfire, have we confirmed the Hydra patches we maintain in this repo still make sense on this rev?

Would it be simpler to keep the same rev, and create an additional patch for submodule support?

disassembler:
This is still pointing to our fork, with 2 additional commits on it. Here's a diff:
```

deepfire:
@jmitchell, there are no patches involved on our fork of Hydra anymore, thankfully!

PR #472, opened 2018-11-08
Module for a Bazel REST cache based on NGinx. [WIP]

jbgi:
After that, not too sure how/where this should be deployed? using packet?

### DIFF
--- a/deployments/build-infrastructure-env-poc.nix
+++ b/deployments/build-infrastructure-env-poc.nix
@@ -1,0 +1,24 @@
+{ ... }:
+
+with (import ./../lib.nix);
+{
+  network.description = "IOHK build infrastructure POC";
+
+  defaults = {
+    services.dd-agent.tags = ["env:poc" ];
+  };
+
+  bazel-cache = { config, pkgs, resources, ... }: let
+    hostName = "bazel-cache.aws.iohkdev.io";
+  in {
+    imports = [
+      ../modules/common.nix
+      ../modules/nginx-bazel-cache-service.nix
+    ];
+    services.nginx-bazel-cache = {
+      enable = true;
+      publicHost =  hostName;
+      writePassword = readFile ../static/bazel-cache-write-password.secret;
+    };
+  };
+}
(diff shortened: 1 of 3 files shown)


disassembler:
I would prefer this to be tested as part of the proof of concept in a separate deployment from the staging deployer. I'd recommend AWS instead of packet.net (it's cheaper and we really don't need the performance gains of bare metal here). It's also a lot easier to use with nixops as our packet.net module isn't used yet.

iohk/common/Types.hs
```diff
@@ -140,6 +141,7 @@ data Environment
   | Staging
   | Testnet
   | Development
+  | Poc
```
disassembler: I don't think we should add the POC to io. In my opinion, the POC is temporary and will never be merged into master.

scripts/ci.sh
```diff
@@ -16,8 +16,9 @@ WITH_REPORT_SERVER=true
 WITH_INFRA_PRODUCTION=true
 WITH_INFRA_STAGING=true
 WITH_BENCHMARK=true
+WITH_BUILDINFRA_POC=true
```
disassembler: Lets not touch the ci.sh test script for this. We can move this into INFRA_PRODUCTION if we decide we want to go this route.

topology-poc.yaml
```diff
@@ -0,0 +1 @@
+nodes:
```
disassembler: we shouldn't need a topology for poc. Topology files are for cardano-node which this doesn't need.

iohk/common/Constants.hs
```diff
@@ -145,6 +146,13 @@ envSettings env =
                                , (Explorer,       All, "deployments/cardano-explorer-env-development.nix")
                                , (ReportServer,   All, "deployments/report-server-env-development.nix")
                                ] <> deplAgnosticFiles}
+    Poc      -> EnvSettings
```
disassembler: lets not touch this file either.

PR #498, opened 2018-12-03
[DEVOPS-1165] Updates & fixes arising from last update proposal

rvl:
Various little things needed updates or fixes.

### DIFF
--- a/docs/Daedalus-Installer-Update-Procedure.md
+++ b/docs/Daedalus-Installer-Update-Procedure.md
@@ -239,19 +239,14 @@ network where the update will be proposed.
 | testnet |                 `0.1.0` |
 
 
-## Voter index table
+## Voter index
 
-For whatever reason, the update proposal procedure requires a voter
-index. This number is supposed to denote the person who submitted the
-proposal.
+The update proposal procedure requires a voter index which selects the
+secret key used to sign the update proposal.
 
-| `voterIndex` | Person      |
-| ------------:| ----------- |
-|            1 | Domen       |
-|            2 | Serge       |
-
-**Note:** The `voterIndex` needs to be *≥ 0 and ≤ 2*, otherwise you
-will receive the error `Prelude.!!: index too large`.
+**Note:** The `voterIndex` needs to be *≥ 0 and < 7* (where 7 is the
+number of core nodes), otherwise you will receive the error
+`Prelude.!!: index too large`.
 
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -46,12 +46,18 @@ To hack on the `iohk-ops` tool, use
...
-        runhaskell -iiohk iohk/iohk-ops.hs "$@"
+        cabal exec iohk-ops -- "$@"
     }
     [nix-shell:~/iohk/iohk-ops]$ io --help
 
 This will provide a Haskell environment where you can use `io` to run
-the script or `ghci` for development.
+the script or `ghci` for development:
+
+    [nix-shell:~/iohk/iohk-ops]$ ghci -iiohk/common
```
cleverca22: you can also use `cabal repl` or `runhaskell Setup.hs repl` to get ghci against a cabal project
i also often do `ghcid -c "runhaskell Setup.hs repl"` to get constant updates about errors

shell.nix
```diff
@@ -16,10 +16,10 @@ let
   ioSelfBuild = pkgs.lib.overrideDerivation iohk-ops.env (drv: {
     shellHook = ''
       function io {
-        runhaskell -iiohk iohk/iohk-ops.hs "$@"
+        cabal exec iohk-ops -- "$@"
       }
       function ghcid-io {
-        ${pkgs.haskellPackages.ghcid}/bin/ghcid -c "ghci -iiohk iohk/iohk-ops.hs"
+        ${pkgs.haskellPackages.ghcid}/bin/ghcid -c "ghci -iiohk -iiohk/common iohk/iohk-ops.hs"
```
cleverca22: `ghcid -c "runhaskell Setup.hs repl"` again
although, there is the pain-point that repl has to be ran in the iohk dir, and the io util wants to not be ran there (but i suspect `-C ../config.yaml` can help)

README.md
rvl: Yes I would prefer `cabal repl` but it needs to be `ghci` because of the iohk-ops working directory thing. If I type `:cd` in `cabal repl` then it wants to reload all the modules.

shell.nix
rvl: Unfortunately `-C ../config.yaml` doesn't work because it won't look up topology files relative to the config file. We should fix that, but not here.

rvl:
Thanks.

bors r+

PR #507, opened 2018-12-13
[DEVOPS-1182] Build timeouts for nix-darwin

rvl:


nix-darwin/modules/basics.nix
```diff
@@ -38,6 +38,13 @@ in {
   nix.extraOptions = ''
     gc-keep-derivations = true
     gc-keep-outputs = true
+
+    # Max of 4 hours for building any given derivation on macOS.
+    # See ../modules/hydra-slave.nix for Linux setting
+    timeout = ${toString (3600 * 4)}
```
angerman: This will prevent cross ghcs from being built on the mac builders I assume. I've seen them take 7-8hs.

PR #511, opened 2018-12-19
[DEVOPS-1179] jobsets: Clean up and add jobset for cardano-wallet

rvl:
* Adds jobsets for the cardano-wallet repo.
* The file needed to be cleaned up because there was code duplication for every project.

### Testing steps

Build the jobset spec with example data on the master branch and PR branch, then compare results with a diff program.

    cat $(nix-build jobsets/default.nix --no-out-link) | jq . > spec.json

### DIFF
--- a/jobsets/default.nix
+++ b/jobsets/default.nix
@@ -91,7 +91,7 @@ let
     };
 
     cardano-wallet = {
-      description = "Cardano Wallet";
+      description = "Cardano Wallet Backend";
       url = "https://github.com/input-output-hk/cardano-wallet.git";
       branch = "develop";
       prs = walletPrsJSON;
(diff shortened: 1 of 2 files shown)


jobsets/default.nix
```diff
@@ -57,6 +134,82 @@ let
...
+  in
+    nameValuePair name jobset;
+
+  # Make jobsets for extra project branches (e.g. release branches)
+  mkJobsetBranches = { name, description, url, input }:
+    mapAttrsToList (suffix: branch:
+      mkJobset { name = "${name}-${suffix}"; inherit description url input branch; });
+
+  # Make a jobset for a GitHub PRs
+  mkJobsetPR = { name, description, input, modifier }: num: info: {
```
jbgi: `description` is not use. Should it be optional with default value `"PR ${num}: ${info.title}"`?

jobsets/default.nix
```diff
@@ -1,34 +1,111 @@
...
 , ouroborosNetworkPrsJSON ? ./simple-pr-dummy.json
 , chainPrsJSON ? ./simple-pr-dummy.json
-, nixpkgs ? <nixpkgs>
-, declInput ? {}
-, handleCardanoPRs ? true
+, walletPrsJSON ? ./simple-pr-dummy.json
 }:
 
-# Followed by https://github.com/NixOS/hydra/pull/418/files
+let pkgs = import nixpkgs {}; in
```
cleverca22: we may want to update `spec.json`, this is importing a nixpkgs from the 16.09 channel

PR #525, opened 2019-02-01
jobset: add iohk-nix

angerman:
tested with
```
jq . < $(nix-build --no-out-link jobsets/default.nix)
```


angerman:
> The jobset file `iohk-nix/release.nix` should accept an `iohk-nix` argument for the source because hydra will be passing it.

Ahh! I think I'm understanding something I've been wondering about a lot!

cleverca22:
you can also use `./foo` in the `release.nix` to refer to itself
the main use for the `iohk-nix` arg in `release.nix` is to get the git revision that hydra has cloned, since `.git` will be missing

PR #526, opened 2019-02-05
mantis service [WIP]

jbgi:


modules/mantis-service.nix
```diff
@@ -346,7 +345,7 @@ in
     };
 
     monetaryPolicyRewardReductionRate = mkOption {
-      type = localLib.nixpkgs.lib.types.float;
+      type = lib.types.float;
```
deepfire: Are you sure that works with 18.09?  I had some error when doing a similar change recently..

PR #529, opened 2019-02-13
[DEVOPS-1216] 18.09 fixes (hydra/ntp/extra-statsd)

disassembler:


modules/ntp_fix.nix
```diff
@@ -1,8 +1,8 @@
 let
   overlay = self: super: {
-    ntp = super.ntp.overrideAttrs (drv: {
-      patches = drv.patches or [] ++ [ ./openat.patch ];
-    });
+    #ntp = super.ntp.overrideAttrs (drv: {
+    #  patches = drv.patches or [] ++ [ ./openat.patch ];
+    #});
```
deepfire: :100:

shell.nix
```diff
@@ -7,7 +7,7 @@ let
   iohk-ops = iohkpkgs.iohk-ops;
   justIo = pkgs.mkShell {
     name = "io";
-    buildInputs = with pkgs; [ iohk-ops terraform_0_11 iohkpkgs.nixops cabal-install ];
+    buildInputs = with pkgs; [ iohk-ops terraform_0_11 iohkpkgs.nixops ];
```
deepfire: @disassembler, is `cabal-install` broken?  It's useful to have it..

disassembler: as discussed, this can be added back in if needed in another PR, but cabal-install was causing a full rebuild of GHC that we couldn't wait a few hours for to deploy the deployer.

PR #532, opened 2019-02-20
Add: Hydra configuration for fm-ledger-rules repository.

nc6:


nc6:
@disassembler What's the motivation for using bors? We can do so, but it makes certain things annoying (e.g. no squash merges)

dnadales:
I would also prefer not using `bors` at the moment. We're still a small team and the costs do not overweight the benefits IMHO...

BTW, we have a buildkite job that seems to be stale.

PR #534, opened 2019-02-26
infra: Add GitHub deploy keys to buildkite-agent

rvl:
Needed for input-output-hk/haskell.nix#56.

Before deploying, create these keys on the infra deployer:
```
ssh-keygen -t rsa -f static/buildkite-hackage-ssh -C buildkite-agent@hackage.nix
ssh-keygen -t rsa -f static/buildkite-stackage-ssh -C buildkite-agent@stackage.nix
```

After deploying, add these as deploy keys to the relevant github repos.

### DIFF
--- a/modules/buildkite-agent.nix
+++ b/modules/buildkite-agent.nix
@@ -57,6 +57,7 @@ in
       user    = "buildkite-agent";
       permissions = "0770";
     };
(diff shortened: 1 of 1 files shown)


rvl:
Required jobs should be fixed now.

bors r+

PR #538, opened 2019-03-05
Add: Report status to fm-ledger-rules repo.

nc6:


disassembler:
can you please change serokell -> Cardano? We're trying to disable the serokell project (and they're identical).

dnadales:
> can you please change serokell -> Cardano? We're trying to disable the serokell project (and they're identical).

Done. Thanks for the review @disassembler

dnadales:
@cleverca22 @disassembler `fm-ledger-rules` was renamed to `cardano-ledger-specs`, I guess this will require further changes. Could you point us to what we should change?

CC: @ruhatch

disassembler:
The names in jobsets/default.nix and jobsets/spec.json

dnadales:
> The names in jobsets/default.nix and jobsets/spec.json

Done. Thanks!

PR #549, opened 2019-03-20
Buildkite: fix NIX_PATH and add deployment key for haskell.nix

rvl:
Fix problems found when doing input-output-hk/haskell.nix#93.

### DIFF
--- a/modules/buildkite-agent.nix
+++ b/modules/buildkite-agent.nix
@@ -96,6 +96,13 @@ in
       user    = "buildkite-agent";
     };
 
+    # GitHub deploy key for input-output-hk/haskell.nix
+    # (used to update gh-pages documentation)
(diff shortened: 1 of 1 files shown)


rvl:
> Is there anyway one could exfil this key?

Yes, a buildkite build could `cat` the ssh key file and it would appear in the logs.

angerman:
> > Is there anyway one could exfil this key?
> 
> Yes, a buildkite build could `cat` the ssh key file and it would appear in the logs.

Should we prevent buildkite to run on untrusted PRs then?

rvl:
> Should we prevent buildkite to run on untrusted PRs then?

Yes. Already, Buildkite will not run PR builds from users who don't have repo write access.

PR #553, opened 2019-03-21
[DEVOPS-1252] begin migration to the Cardano project in hydra

cleverca22:
this initial PR configures hydra to report the `required` job on both the `Cardano` and `serokell` project

that will allow smoothly transitioning the required CI checks in each repo over

### DIFF
--- a/modules/hydra-master-main.nix
+++ b/modules/hydra-master-main.nix
@@ -97,7 +97,7 @@ in {
 
       ${mkStatusBlocks [
         { jobset = "iohk-ops"; inputs = "jobsets"; }
-        { jobset = "cardano-ledger"; inputs = "chain"; }
+        { jobset = "cardano-ledger"; inputs = "cardano-ledger"; }
         { jobset = "cardano-ledger-specs"; inputs = "cardano-ledger-specs"; }
         { jobset = "cardano"; inputs = "cardano"; }
         { jobset = "plutus"; inputs = "plutus"; }
--- a/modules/hydra-master-main.nix
+++ b/modules/hydra-master-main.nix
@@ -97,8 +97,8 @@ in {
 
       ${mkStatusBlocks [
         { jobset = "iohk-ops"; inputs = "jobsets"; }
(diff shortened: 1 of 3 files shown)


modules/hydra-master-main.nix
```diff
@@ -79,30 +94,19 @@ in {
...
-      <githubstatus>
-        jobs = serokell:cardano.*:required
-        inputs = cardano
-        excludeBuildFromContext = 1
-        useShortContext = 1
-      </githubstatus>
+
+      ${mkStatusBlocks [
+        { jobset = "iohk-ops"; inputs = "jobsets"; }
+        { jobset = "cardano-ledger"; inputs = "chain"; }
```
jbgi: btw, what in the role of the `inputs` attributes? (any why `chain`?)

cleverca22: `inputs` is the name of the input from the hydra config, and i was just basing it on the old xml-ish config
but it looks like i need to include the rename from the other pr

modules/hydra-master-main.nix
```diff
@@ -79,30 +94,19 @@ in {
...
-        inputs = cardano
-        excludeBuildFromContext = 1
-        useShortContext = 1
-      </githubstatus>
+
+      ${mkStatusBlocks [
+        { jobset = "iohk-ops"; inputs = "jobsets"; }
+        { jobset = "cardano-ledger"; inputs = "cardano-ledger"; }
+        { jobset = "cardano-ledger-specs"; inputs = "cardano-ledger-specs"; }
+        { jobset = "cardano"; inputs = "cardano"; }
```
jbgi: this jobsset, `cardano.*` will also match jobs from `cardano-ledger.*` (which in turn matches jobs from `cardano-ledger-specs.*` ). Would that create multiple notifications?

cleverca22: the notify plugin in hydra will stop at the first regex to match, and ignore all others that follow

jbgi: ok thanks! so we need `cardano-ledger-specs` before `cardano-ledger`, then?

PR #558, opened 2019-03-29
New monitoring stack based on prometheus. Fix #543.

### LINKED ISSUE #543: # Prometheus Integration
Need to have a monitoring server with the following for each deployment:

* prometheus
* alertmanager
* grafana
* all services behind oauth_proxy
* integration with pagerduty
* statsd exporter for metrics for applications providing metrics
* node exporter on all nodes in deployment
* nginx /w vts monitoring on any nginx hosts (including monitoring host)
* deadmansnitch to alert if prometheus server is down (PD integration as well)

jbgi:


modules/monitoring-exporters.nix
```diff
@@ -0,0 +1,106 @@
...
+        script = ''
+          ${pkgs.prometheus-statsd-exporter}/bin/statsd_bridge -statsd.listen-address ":8125" -web.listen-address ":9102" -statsd.add-suffix=false
+        '';
+      };
+
+      services = {
+        prometheus.exporters.node = {
+          enable = true;
+          enabledCollectors = [
+            "systemd"
```
disassembler: lets add ntp

disassembler: we may also want processes

deployments/monitoring-env-production.nix
```diff
@@ -0,0 +1,10 @@
+{ ... }:
+{
+  monitoring = { ... }:
```
cleverca22: i believe it is equally valid to make the entire file just:
```
{
  monitoring = ../modules/production.nix;
}
```
but i also want to see `require` used more in iohk-ops, so it would be:
```
{
  require = [ ./monitoring.nix ];
  monitoring = ../modules/production.nix;
}
```
then `deployment/monitoring.nix` doesnt have to be passed in at `nixops modify` time, making the haskell side a bit simpler

deployments/monitoring.nix
```diff
@@ -0,0 +1,57 @@
...
+      ../modules/network-wide.nix
+      ../modules/monitoring-services.nix
+    ];
+
+    global = {
+      organisation = monitoring.org;
+      dnsHostname  = mkForce "monitoring";
+    };
+  } // (let
+    hostList = lib.mapAttrsToList
```
cleverca22: i would move this up a level, and just not use `//`
```
monitoring = { config, lib, pkgs, resources, nodes, ... }:
let
 key=val;
in {
imports .....
```

mainnet.yaml
```diff
@@ -29,9 +29,12 @@ files:
 - deployments/cardano-explorer-env-production.nix
 - deployments/report-server.nix
 - deployments/report-server-env-production.nix
+- deployments/monitoring.nix
```
cleverca22: now that `require` is in use, i dont think `monitoring.nix` has to be listed here anymore, it gets pulled in by `monitoring-env-staging.nix`
also, this is mainnet, why is it loading `monitoring-env-staging.nix`?

deepfire:
I only have one question -- do you think `monitoring-services.nix` can be at least partially upstreamable?

PR #563, opened 2019-04-01
modules/cardano-explorer:  t2.large -> t3.xlarge (2x RAM, newer system)

deepfire:
This is a sweeping change that affects Explorer deployment on all environments.


deepfire:
This is a sweeping change that affects Explorer deployment on all environments.

disassembler:
I think we don't need this since it was just a temporary bump, right @deepfire @jbgi?

deepfire:
Right, obsolete now.

PR #565, opened 2019-04-04
Use the .png version of the iohk logo in Hydra.

kreisys:
It looks better with a transparent background.


cleverca22:
Screenshot_2019-04-04_19-17-37

looks a bit worse once its down-scaled to fit in the UI

kreisys:
That's... strange. Doesn't look at all like that on my end 🤷‍♂️ 

Screen Shot 2019-04-04 at 6 19 24 PM

cleverca22:
it may depend on resolution and DPI

kreisys:
What’s bors and why ain’t it returning no status 🤨

disassembler:
bors ping

PR #576, opened 2019-05-01
Rebase `goguen-ala-cardano` onto `master`

kreisys:
in preparation for merge.


.gac.defaults
```diff
@@ -0,0 +1,4 @@
+CLUSTER_KIND=mantis
+CLUSTER_TYPE=mantis
+OPS_REPO='git@github.com:input-output-hk/iohk-ops'
+DEFAULT_OPS_BRANCH='goguen-ala-cardano'
```
disassembler: should be master. Do we want to keep default to mantis for kind/type? Probably?

configs/default.nix
```diff
@@ -0,0 +1,41 @@
...
+{
+  node = {
+    ## Default values for organisation and region.
+    org                        = "IOHK";
+    region                     = "eu-central-1";
+  };
+
+  cluster = {
+    ## Do not allocate hosted zones by default.
+    hostedZone                 = "project42.iohkdev.io";
```
disassembler: change this to something else

configs/goguen-ala-cardano.nix
```diff
@@ -0,0 +1,12 @@
+{
+  node = {
+    org                  = "IOHK";
+    region               = "eu-central-1";
+  };
+
+  cluster = {
+    hostedZone           = "project42.iohkdev.io";
```
disassembler: change to something else

deployments/monitoring.nix
```diff
@@ -5,11 +5,11 @@ let
 in
 
 {
-  monitoring = { config, lib, pkgs, resources, nodes, ... }:
+  monitoring = { config, lib, pkgs, resources, nodes, ... }:
```
disassembler: don't bring back bad whitespace in this file

deployments/deployer/deployer-dev-access.nix
```diff
@@ -0,0 +1,5 @@
+with (import ../../lib.nix);
```
cleverca22: parens arent really needed

deployments/hydra/hydra-aws.nix
```diff
@@ -0,0 +1,32 @@
...
+
+{
+  hydra.imports = [ <module/hydra-aws.nix> ];
+
+  resources = {
+    elasticIPs.hydra-ip = { inherit region accessKeyId; };
+    ec2SecurityGroups = {
+      "allow-hydra-${region}-${org}" = { resources, ...}: {
+        inherit region accessKeyId;
+        description = "SSH";
```
cleverca22: description claims ssh, but rules allow ssh, http, and https

deployments/mantis-deployer.nix
```diff
@@ -0,0 +1,58 @@
+{ ... }:
+
+let lib = import ../lib.nix; in
+with lib;
+with (import ../lib/ssh-keys.nix { inherit lib; });
```
cleverca22: redundant parens again

lib/ssh-keys.nix
```diff
@@ -19,6 +19,7 @@
...
     charles        = [ "ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBJBRhsdeDa6/lvaEOWKPLFJnc2aNsBE+aCiWh+PTVq5uAEPWRh6HdJfj90jXuLGprdRanFjV0vAO2xL3gB5wRxE= cmorgan@Marius.local" ];
+    felix          = [ "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCvEEepJffFzPZDFTEUFhADszQ+9QfoFTyYod7NTgMDoVnnFxxGNbjlvbA6EUzet5TlYtuYataFKQsC6gJdesRH8FrIQNbiJUP1AWQHVsstlqUMCrDXgDMntUVaWY7hiUxYbFncO95It6hzC2CGodwxywMQnLTci/Prk8Y4zSrdkUDZdmZsajJhBNyiatd6UUujkic8yRXCMmnG/DNIzNV09d15wm7IRjkKWJAHmKlUu5srMr5BX153RsY7FbIUCkcCmq//hLXQDLOeWVi0I2PdzywJAJ8yw36tRjeAhewNHWjROipQGK5czEr/ElU6agdpWqzw7PYyfSAnY73QIAT7" "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQClF9ko5u4zf0CEvleEeRbo9r6BMNgXEGO/rDNZOEHcKxVaeIi+/xF6ZQ5MZbcmH08lswq32hb1XwXg7Gk+ofUdEvCD/kC/vJijt7IFkardy6BNOSWQJLEf6/BpL3LzDQhi7iZXPF46VYoPVGHBh8fKQaAtOCrhbf/8JutfTwCglEztjoiQxY5b8OMfntjBSl6TJwZPJAoQllbJJz9q90sBetvqx6Y08eqIzsSZw6pznpvivRR+TSKU0EkVYS2y2zBAvPK6oyunj5zi01/FACT+Qn70dUkumZAvcPssbl0hCs/xDLgEL6hCEvoszodyMYVn7HS0KwfUlfiGdNUOFHIl" ];
```
disassembler: felix shouldn't be in devopsKeys

lib/ssh-keys.nix
```diff
@@ -47,6 +48,7 @@
...
     robcohen       = [ "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCQkixVdLYB664tY3foVWa1iCIamHTB8E1kgRuNijMLY9h9+S7g8pZmcfxwwl3I0/xDVEKBb3wfhZLRJ2SpbhPuBaoODiYY14+ZpTwJszq0+AustGqiaUxS7HtOF6oTcVtxSbRHBC4zS1ibK0XsOTJAaD713RlunE1TSGZGkhgHF1xUNGE8DWYrPnS0DGx3jJkVqAMXSPVyyJhM+DdUrfSLxqiRuqWKJMtoZ/kHEcA+EvSp34x+EhHuljHzIbqnYbw87vO/WNgDzxmpGWtnacvffnVR1OWjVi+G8+hgILyGJBSvRQbmkhwR3SFUlPr6rzONXTW/opAj8gn3ZbJRIdGLEJmd6XuPEGG5OkdIWv0Jlu0ZRFHWj6PNxGXYGCaS06D1rKv+3aLREzO106i1odYl2EnIcWU5LSL28csTk2pJs8xpH+M8HGeeTXxoxFUcQOYVGwHPh4RUe9pUc0hWayo2uajZ+Okms34uavps2yLTzXyE3E76T+al4W1nyOG9HruAHumvDlWbe9oJdJ4mWBOvJModR6Vt8Gz2XFOf5CDcvWUlc14Qk80mC61eV3hRZi8CT12AtcQjE0bG/Tv+LkDNbNyUxU9LsSkPPcADvzp6vXOEr+uMQxXbawUwCtOgPhKN3pmoqRQ2YAfxrESXYFZ6YvEVYjzSLawbDEM/BlG+Nw== Minx"];
+    alanv          = [ "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCvQkw2IYLIOiQysBJn5UuHkd7mgNkY0Q/TPP2uCPUGBZZGSIop4BzCnnPOYKn+zaMt4bLmdUuWHDs7+XKymV5RlmeMrptwIAL9XWX3d1AnWMCRgaE9ByXkEIMvMeKxUq3y2pY0e9zZLWsJvtZNzeTy227NQYYOxgx/IXdCacKdPEKENWBq6sPehP+9q+0uKW6zVUXHsCUhBXlE/FCwdh1HKoOSKyFYsvu8tIRF5u0crHQ1zDANw9ASRhmYoxvHgfEBY3GiwywTh/JHRuMS5CyD4PeKSOmYq6BALZVyMun0Ol8+XcOh63ljOlfZJGR/xWDaE0lFSuwlkG2UmGI7jLIJ alverbner@gmail.com" ];
```
disassembler: should probably have a separate group in subrepos utilizing this for those specific developers.

modules/common.nix
```diff
@@ -63,11 +63,13 @@ with (import ./../lib.nix);
     nixPath = [ "nixpkgs=/run/current-system/nixpkgs" ];
 
     # use our hydra builds
-    trustedBinaryCaches = [ "https://cache.nixos.org" "https://hydra.iohk.io" ];
+    # XXX: if there are any objections for adding mantis-hydra (potential security issues), let's discuss them
+    trustedBinaryCaches = [ "https://cache.nixos.org" "https://hydra.iohk.io" "https://hydra.project42.iohkdev.io/" ];
```
disassembler: remove this change. Extend it in repos using this as a module.

modules/common.nix
```diff
@@ -63,11 +63,13 @@ with (import ./../lib.nix);
...
     binaryCaches = trustedBinaryCaches;
     binaryCachePublicKeys = [
       "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY="
       "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ="
+      "hydra.iohk.io-1:chtUuea0mkt7j3Q3ESvfJUeqTNNPspSO//Yl6O00p/Y="
```
disassembler: same here

modules/deployer-bootstrap.nix
```diff
@@ -0,0 +1,63 @@
...
+let
+  username = "deployer";
+
+in {
+    imports = [
+      <nixpkgs/nixos/modules/virtualisation/amazon-image.nix>
+    ];
+    ec2.hvm = true;
+
+    users.users.${username} = {
```
disassembler: I think we should reuse keys from lib.nix

modules/exporer-remix-solidity-services.nix
```diff
@@ -0,0 +1,87 @@
+{ pkgs, lib, nodes, options, config, ... }:
```
disassembler: typo in filename? exporer? should be explorer maybe?

modules/hydra-base.nix
```diff
@@ -0,0 +1,8 @@
+{ config, ... }: {
+  imports = [
+    ./common.nix
+    ./hydra-master-common.nix
+  ];
+
+  services.dd-agent.tags = [ "role:hydra" "depl:${config.deployment.name}" ];
```
disassembler: Don't use DD

deployments/mantis-deployer.nix
```diff
@@ -0,0 +1,58 @@
...
+      "https://cache.nixos.org"
+      "https://hydra.iohk.io" "https://mantis-hydra.aws.iohkdev.io"
+    ];
+    nix.binaryCachePublicKeys = mkForce [
+      "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY="
+      "hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ="
+      "hydra.iohk.io-1:E8yDJv2SBXM6PQPVbhCWK7VvitistFYSH2u3AuwCiu4="
+    ];
+
+    services.dd-agent.tags = ["env:production" "depl:${config.deployment.name}" "role:deployer"];
```
disassembler: don't use DD

modules/deployer-base.nix
```diff
@@ -1,48 +1,27 @@
...
 
-  environment.systemPackages = (with iohk-pkgs; [
-    iohk-ops
-    terraform
-    mfa
-  ]) ++ (with pkgs; [
-    psmisc
-    gnupg
-    nixops
+  services.dd-agent.tags = [ "role:deployer" "depl:${config.deployment.name}" ];
```
disassembler: don't use DD

modules/hydra-master-common.nix
```diff
@@ -8,7 +8,7 @@ in {
   nix = {
     extraOptions = ''
       auto-optimise-store = true
-      allowed-uris = https://github.com/NixOS/nixpkgs/archive https://github.com/input-output-hk
+      allowed-uris = https://github.com/NixOS/nixpkgs/archive https://github.com/input-output-hk ssh://git@github.com/input-output-hk
```
disassembler: I don't think we should allow ssh by default. We should override where needed. We don't want to accidentally pull something from private repos.

static/faucet-addresses.nix
```diff
@@ -0,0 +1,3 @@
+{
+faucet-a = "6413339a3d72e2b4f4e5044854dc0f9607591ae2";
+{
```
disassembler: static files should not be committed, and nix is invalid

static/mallet.json
```diff
@@ -0,0 +1,81 @@
+{
```
disassembler: static files shouldn't be committed

static/mantis-a-0.key
```diff
@@ -0,0 +1,2 @@
+009872b6d988d4d208e8ccf89ec7da99515b4ec6bc86f75ab9d9ae1c6a9c435380
```
disassembler: static files should not be committed

static/mantis-a-0.key
```diff
@@ -0,0 +1,2 @@
...
+2cb6e52366190095b79c7ef13180523c3fe848f71d991f6d4cef4ae45aae94c59ae1ae2a2ddecfa94762cf4f9fb7038942edf9f74a7967362cee8bc498f4c036
```
disassembler: static files should not be committed

static/mantis-a-1.key
```diff
@@ -0,0 +1,2 @@
+67cdbe53f16cafa1fd2a0f67ddf97b96bf9e33a2d7cb0d82c9ab0bab6f0eae7e
```
disassembler: static files should not be committed

static/mantis-b-0.key
```diff
@@ -0,0 +1,2 @@
+4dea3263bc456d10c7550507938104f475138249551918e7621bc40a90cd93aa
```
disassembler: static files should not be committed

deployments/monitoring.nix
johnalotoski: Extra whitespace at EOL on 3 lines

static/mantis-b-1.key
```diff
@@ -0,0 +1,2 @@
+009e77a19ad8f19cc771f2af5cf7de2883751437746403691defad800eb8e93f74
```
disassembler: static files should not be committed

static/mantis-c-0.key
```diff
@@ -0,0 +1,2 @@
+00fe472d59f31f8aa33df4d7edd517b774c86625413b5a5b820667d0c8e22ae056
```
disassembler: static files should not be committed

static/mantis-faucet-keygen.conf
```diff
@@ -0,0 +1 @@
+include "application.conf"
```
disassembler: static files should not be committed

static/mantis-node-ids.nix
```diff
@@ -0,0 +1,7 @@
+{
```
disassembler: static files should not be committed

configs/default.nix
kreisys: 😱

lib/ssh-keys.nix
kreisys: makes sense.

modules/deployer-bootstrap.nix
kreisys: Yeah I also pondered that when I saw this. The problem is that this is passed to ec2 instances as `user-data` which is a metadata field (with a limited size) that can be used to pass arbitrary data to the AMI. The NixOS AMI uses this as its `configuration.nix`. Point is that this file ends there alone so it can't just `import ../lib/ssh-keys.nix` so this requires some thought. I had a few ideas, some dumber than others, but most involved more engineering acrobatics than I felt this was worth. tl;dr: I think the simplest solution here would be to simply remove all the keys in favour of a single one: bootstrapping is a one off operation. If I understand correctly how this works, then this config is temporary and its sole purpose is to prepare the instance for NixOps to scp a closure and activate it. Should be pretty straight forward to modify `gac-mkdeployer.sh` to template-in a public key (either provided by the user or generated. probably doesn't matter); question is whether this should be scoped here or attacked separately later.

modules/exporer-remix-solidity-services.nix
kreisys: 👩‍🏫

modules/hydra-master-common.nix
kreisys: This is probably there because someone in another repo needed it though. For some reason the NixOS module doesn't define an option for `allowed-uris` so adding more uris from other modules would require augmenting the module with that option. Shouldn't be hard. Scope it?

static/mantis-b-0.key
kreisys: Removed

static/mantis-c-0.key
kreisys: Removed

static/mantis-b-1.key
kreisys: Removed

static/mantis-faucet-keygen.conf
kreisys: Removed

static/mantis-node-ids.nix
kreisys: Removed.

PS: the following files were already tracked in `master`, so were *NOT* removed:

```
static/id_buildfarm.pub
static/txgen-logging-warnonly.yaml
static/txgen-logging.yaml
```

modules/deployer-bootstrap.nix
deepfire: This is arguably the simplest solution -- the downside cost being people accidentally able/unable to deploy a new deployer (which is independently controlled by AWS).  So the real cost, in the end, is inability of new members to deploy a new deployer, which isn't a terribly frequent situation.

Any other costs I'm missing?

If not, is it worth complicating things in a non-trivial way just to streamline this single case?

lib.nix
```diff
@@ -2,31 +2,115 @@
...
+    oraclejdk8 = super.callPackage ./goguen/jdk-override/jdk8cpu-linux.nix {
+      installjdk = true;
+      pluginSupport = false;
+      licenseAccepted = true;
+    };
+  };
+  graalvm8 = (import iohkNixGoguen.nixpkgs javaOverrideNixpkgsConfig).graalvm8;
 in lib // (rec {
+  inherit (iohkNix) nixpkgs;
+  inherit mkIohkNix fetchProjectPackages pkgs graalvm8;
```
deepfire: Ideally, we wouldn't want to define app-specific things in the common path, but instead downstream.

Maybe not for now, since we want to expedite this PR.

deployments/testnet-deployer-aws.nix
```diff
@@ -1,18 +0,0 @@
-{ ... }:
```
deepfire: I'm worried by the changes introduced to deployers.

The situation is worsened by the fact we do not have a single deployer deployment mechanism between mainnet/staging and testnet.

I'd like to have at least the staging deployer tested:  deploy one before this PR, and then redeploy it with the PR.

Or maybe it's simpler to make sure there's no change at all to the current deployers..

Oh, the temptation for refactoring.. : -/

modules/hydra-master-common.nix
kreisys: For now we're just going to `mkForce`-override this where it's needed.

PR #578, opened 2019-05-06
Add tools job

angerman:
`tools` will contain compilers, and tools from the @input-output-hk/tools.  We'll also have sample applications.

I just need to figure out how to merge multiple `release.nix` into a top-level `release.nix` expression easily.


angerman:
@cleverca22 can you help me get this one over the line? I believe there's some deeper hydra issue here I don't understand.

PR #580, opened 2019-05-08
fix the hydra eval

cleverca22:
the new nixpkgs-src.json refers to a git commit that does not exist upstream

this results in hydra failing to fetch the specified commit from upstream when supplying <nixpkgs> to all jobsets

### DIFF
--- a/jobsets/default.nix
+++ b/jobsets/default.nix
@@ -165,7 +165,7 @@ let
     schedulingshares = 42;
     checkinterval = 60;
     inputs = {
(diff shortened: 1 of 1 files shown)


jbgi:
I think the commit that were necessary for the monitoring work as been upstreamed to nixos-19.03 so using tip of that channel in `nixpkgs-src.json` should also be fine. Going to test this in staging.

cleverca22:
yeah, switching back to an upstream rev is also a valid fix

PR #584, opened 2019-05-14
Avoid reference on globals from monitoring-exporters.nix

jbgi:
so that it can be used outside cardano specific deployments.

### DIFF
--- a/modules/staging.nix
+++ b/modules/staging.nix
@@ -14,6 +14,8 @@ with import ./../lib.nix;
 
       # DEVOPS-64: disable log bursting
       journald.rateLimitBurst    = 0;
+
+      monitoring-exporters.graylogHost = "${config.deployment.arguments.globals.monitoringNV.name}-ip:5044";
     };
 
   };
--- a/modules/devops.nix
+++ b/modules/devops.nix
@@ -14,6 +14,8 @@ with import ./../lib.nix;
 
       # DEVOPS-64: disable log bursting
       journald.rateLimitBurst    = 0;
+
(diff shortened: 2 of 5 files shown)


modules/cardano-staging.nix
```diff
@@ -8,5 +8,7 @@ params:
     ./papertrail.nix
   ];
   
+  services.monitoring-exporters.graylogHost = "${config.deployment.arguments.globals.monitoringNV.name}-ip:5044";
```
johnalotoski: If we move this assignment one abstraction level higher, we'll have less files to need to include it in.  The iohk/common/Constants.hs file shows monitoring configured for the following environments at present: devops, testnet, staging and production.  If we then move this assignment out of this file and instead to each of those respective environment files (modules/{devops,testnet,staging,production}.nix), I think we should be covered.  Would that work for you?

modules/staging.nix
```diff
@@ -14,6 +14,8 @@ with import ./../lib.nix;
 
       # DEVOPS-64: disable log bursting
       journald.rateLimitBurst    = 0;
+
+      monitoring-exporters.graylogHost = "${config.deployment.arguments.globals.monitoringNV.name}-ip:5044";
```
jbgi: @johnalotoski, correct me if that is not what you suggested.

johnalotoski: @jbgi, this tests good.  It looks to me like the same single line addition needs to be made to testnet.nix and production.nix also for logging to work in those environments.  Additional conditional checks will be added shortly in another PR.

PR #587, opened 2019-05-16
Use options for graylog cluster and root secrets.

jbgi:
For better integration in atala where the static folder is now where the previous code was expecting it to be.


deployments/configurator.nix
```diff
@@ -11,6 +11,10 @@ with (import <nixpkgs> {}).lib;
     # for NixOS configs however, we need to make a distinction between defaults and and user-supplied parameters.
     imports = [ <module/configurator.nix> ];
 
+    nixpkgs.overlays = [
+      (import <overlay/monitoring-exporters.nix>)
+    ];
+
```
cleverca22: i have since added that to `modules/monitoring-exporters.nix`, so if your using the existing monitoring stuff, you dont need to add the overlay yourself

PR #588, opened 2019-05-17
Fix `gac genkey`

kreisys:
which I have erroneously broken while trying to satisfy Shellcheck


NinjaTrappeur:
Tested this fix. It indeed fixes this quote issue.

I cannot create a positive review in this repository though.

PR #590, opened 2019-05-17
Fix not being able to disable graylog

kreisys:
due to the systemd services `graylog-preload` being defined regardless of whether graylog is enabled or not (... which causes it to fail, crash and burn). I need to be able to deploy without graylog simply because graylog is currently out of scope for the PR I'm trying to merge (and I require some changes that were committed on top of the graylog changes).


NinjaTrappeur:
Fixes the problem for me.

PR #594, opened 2019-05-21
Support for new node implementations

jbgi:


disassembler:
Looks good to me. I'd like to see a review from @cleverca22 as well.

PR #595, opened 2019-05-23
initial version of wireguard on the monitoring-devops deployment

cleverca22:


deployments/monitoring-env-devops.nix
```diff
@@ -1,11 +1,42 @@
...
+  mkUplink = iohklib.mkMkUplink {
+    central = "192.168.20.1";
+    subnet = "192.168.20";
+    # TODO, `monitoring-ip` will be wrong if monitoring isnt using an elastic ip by that name
+    endpoint = "monitoring-ip:51820";
+  };
+in {
   require = [ ./monitoring.nix ];
-  monitoring = { ... }:
+  a1 = mkUplink 10 ../static/a1.wgprivate;
```
disassembler: this stuff needs to be undone and moved to infra for all packet.net servers.

deployments/monitoring-aws.nix
```diff
@@ -0,0 +1,38 @@
...
+        }];
+      };
+    };
+  };
+  monitoring = { resources, ... }: {
+    deployment.ec2 = {
+      securityGroups = [
+        resources.ec2SecurityGroups."allow-wireguard-in-${region}-${org}"
+        resources.ec2SecurityGroups."allow-to-monitoring-${region}"
+        resources.ec2SecurityGroups."allow-monitoring-static-peers-${region}-${org}"
```
johnalotoski: The wireguard SG replaces the need for this SG.

modules/monitoring-services.nix
```diff
@@ -25,6 +25,11 @@ in {
         default = true;
       };
 
+      enableWireguard = mkOption {
+        type = types.bool;
+        default = false;
```
johnalotoski: True?

modules/monitoring-exporters.nix
```diff
@@ -59,6 +59,11 @@ in {
           Enable papertrail.
         '';
       };
+
+      ownIp = mkOption {
+        type = types.str;
+        description = "the address a remote prometheus node will use to contact this machine";
```
johnalotoski: Could also mention wireguard IP will be used preferentially if enabled.

modules/gen-wireguard-keys.nix
```diff
@@ -0,0 +1,17 @@
...
+  shellHook = ''
+    cd ${toString ../static}
+    umask 077
+    for host in ${toString hosts}; do
+      wg genkey > ''${host}.wgprivate
+      wg pubkey < ''${host}.wgprivate > ''${host}.wgpublic
+    done
+    exit 0
+  '';
+}
```
johnalotoski: I like this.  Maybe it would be good to add similar for graylog and grafana initial cred creation to save new cluster spin up time.

deployments/monitoring-env-devops.nix
cleverca22: this was mainly a testing ground for the concept, and matching code has now been put into the infra deployments
i'm thinking it can be ok to leave wireguard in this devops deployment, for future testing?

modules/monitoring-services.nix
cleverca22: this is enabled in the env files, when wireguard is needed

PR #598, opened 2019-05-24
extra changes to deployments, needed to make it all deploy

cleverca22:


deployments/monitoring-env-production.nix
```diff
@@ -6,6 +6,9 @@
       ../modules/production.nix
     ];
 
+    # warning, this is only valid for infra-staging, need to find a better solution
+    global.dnsDomainname = lib.mkForce "aws.iohkdev.io";
```
johnalotoski: Need to address this to avoid inadvertent DNS overwrites.

PR #600, opened 2019-05-24
gac: split `deployments/configurator.nix`

### LINKED ISSUE #599: gac: `deployments/configurator.nix` has AWS-specific code in it
This file is used by all GAC-based deployments for the purpose of sharing
configuration parameters between NixOS VMs/instances and the Nixops configs that
deploy them. It also contains some definitions of AWS resources. While the
configuration-parameter management aspect of it is needed by all GAC
deployments, the AWS resources are only needed by deployments to AWS instances.
We therefore need to be able to include those parts separately.

kreisys:
into `deployments/gac/{configurator,common-aws}.nix`.

To maintain backward-compatibility I put a placeholder in the original
path (`deployments/configurator.nix`) that `require`s the two new files
and emits a deprecation warning.

This is generally needed for supporting providers other than AWS.

Fixes #599


kreisys:
ugh I opened this from emacs and confused source and target branch. github doesn't let you change this after the fact. closing 🙃

PR #601, opened 2019-05-24
gac: split `deployments/configurator.nix`

### LINKED ISSUE #599: gac: `deployments/configurator.nix` has AWS-specific code in it

kaii-zen:
into `deployments/gac/{configurator,common-aws}.nix`.

To maintain backward-compatibility I put a placeholder in the original
path (`deployments/configurator.nix`) that `require`s the two new files
and emits a deprecation warning.

This is generally needed for supporting providers other than AWS.

Fixes #599

### DIFF
--- a/deployments/gac/common-aws.nix
+++ b/deployments/gac/common-aws.nix
@@ -1,0 +1,21 @@
+{ config, ... }:
+
+{
+  ## Universal resource logic:
+  resources.ec2KeyPairs."cardano-keypair-${config.node.org}-${config.node.region}" = {
+    inherit (config.node) region accessKeyId;
+  };
+
+  resources.ec2SecurityGroups = {
+    "allow-deployer-ssh-${config.node.region}-${config.node.org}" = {
+      _file = ./configurator.nix;
+      inherit (config.node) region accessKeyId;
+      description = "SSH";
+      rules = [{
+        protocol = "tcp"; # TCP
+        fromPort = 22; toPort = 22;
+        sourceIp = config.cluster.deployerIP + "/32";
+      }];
+    };
+  };
+}
--- a/deployments/gac/configurator.nix
+++ b/deployments/gac/configurator.nix
@@ -1,0 +1,19 @@
+{ config , ...  }:
+
+{
+  ## Per-machine defaults:
+  defaults = {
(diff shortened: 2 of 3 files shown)


deployments/gac/configurator.nix
```diff
@@ -0,0 +1,20 @@
+{ config , ...  }:
+
+with (import <nixpkgs> {}).lib;
```
cleverca22: i believe `{ config, lib, ... }:` will also get you lib

deployments/gac/common-aws.nix
```diff
@@ -0,0 +1,20 @@
+{ config, ... }:
+
+{
+  ## Universal resource logic:
+  resources.ec2KeyPairs."cardano-keypair-${config.node.org}-${config.node.region}" = {
+    inherit (config.node) region accessKeyId;
+  };
+
+  resources.ec2SecurityGroups = {
+    "allow-deployer-ssh-${config.node.region}-${config.node.org}" = {
```
cleverca22: with the latest changes in the master branch, you can also do:
`require = [ ./security-groups/allow-deployer-ssh.nix ];`
but that relies on `io` having done `nixops set-args --arg globals 'import ./globals.nix ....'`

deployments/gac/configurator.nix
kreisys: Are you sure? Does Nixops provide a lib argument? I know that `gac` doesn't (but maybe it should)

deployments/gac/common-aws.nix
kreisys: We could do some gymnastics to get around `io`'s `globals` but wouldn't it just make things more difficult to reason about? @deepfire thoughts?

deployments/gac/configurator.nix
kreisys: Actually it doesn't look like this `with` is used at all

PR #603, opened 2019-05-28
Monitoring improvements

cleverca22:


modules/gen-grafana-creds.nix
```diff
@@ -0,0 +1,39 @@
+{ user ? null, password ? null }:
+
+let
+  pkgs = import (import ../fetch-nixpkgs.nix) {};
+in pkgs.stdenv.mkDerivation {
+  name = "gen-grafana-creds";
+  buildInputs = with pkgs; [ pwgen gnused ];
```
johnalotoski: Can drop gnused; this was carried over from gen-graylog-creds.nix, but isn't used here.

modules/monitoring-exporters.nix
```diff
@@ -94,7 +94,7 @@ in {
         requires = [ "network.target" ];
         after = [ "network.target" ];
         script = ''
-          ${pkgs.prometheus-statsd-exporter}/bin/statsd_bridge -statsd.listen-address ":8125" -web.listen-address ":9102" -statsd.add-suffix=false
+          ${pkgs.prometheus-statsd-exporter}/bin/statsd_bridge -statsd.listen-address ":8125" -web.listen-address ":9102" -statsd.add-suffix=false || ${pkgs.prometheus-statsd-exporter}/bin/statsd_exporter --statsd.listen-udp=":8125" --web.listen-address=":9102"
```
johnalotoski: Good idea to cover both possibilities.

PR #606, opened 2019-05-30
Monitoring improvements

johnalotoski:
- Adds nix-shell cred generation message to errors
- Adds graylog pagerduty, auth-sso and slack plugins
- Adds a 1 minute grafana automatic dashboard refresh
- Adds further security group modularization


deployments/security-groups/allow-to-monitoring.nix
```diff
@@ -0,0 +1,25 @@
...
+      _file = ./allow-to-monitoring.nix;
+      accessKeyId = globals.orgAccessKeys.${org};
+      description = "Access to monitoring server";
+      rules = [{
+	protocol = "tcp";
+	fromPort = 80; toPort = 80;
+	sourceIp = "0.0.0.0/0";
+      } {
+	protocol = "tcp";
+	fromPort = 443; toPort = 443;
```
cleverca22: oh, this rule can just be removed
replace it with the existing http(s) rule that also allows 80/443

PR #609, opened 2019-05-31
OSX KVM hydra build slave support

cleverca22:
This PR enables us to run NixOS on apple hardware and spin up VMs using KVM. This lets us move away from 3rd party services giving us virtualized OSX hardware to physical bare metal mac machines.

New features:

* Run wireguard on NixOS host for monitoring, deployment and hydra builds
* Run in a NAT without any open ports so all communication goes through WG tunnel
* Run both hydra and buildkite agents on same machine simultaneously
* Easily test multiple OSX releases
* Rollback any changes to mac guests on restart of service
* CI deployments for mac are done completely using nixops (no more manual steps)
* Prometheus monitoring for both host and guests
* adds cachecache server on host for fast hydra caching to guest machines


lib.nix
```diff
@@ -15,7 +15,10 @@ let
         url = "${spec.url}/archive/${spec.rev}.tar.gz";
         inherit (spec) sha256;
       }) (removeAttrs iohkNixArgs ["iohkNixJsonOverride"]);
-  iohkNix       = mkIohkNix { application = "iohk-ops"; };
+  iohkNix       = mkIohkNix {
+    application = "iohk-ops";
+    nixpkgsJsonOverride = ./nixpkgs-src.json;
```
cleverca22: still need to decide on what to do with this
without that present, half of iohk-ops obeys a different pin, and hydra caching fails
with this present, nixops fails to build, so nix-shell is broken

PR #617, opened 2019-06-10
jobsets: Add cardano-node and cardano-base

rvl:
Corresponding PR is input-output-hk/cardano-node#10

Tested with:

```
jq . < $(nix-build --no-out-link jobsets/default.nix)
```


jobsets/spec.json
```diff
@@ -25,6 +25,7 @@
          ,"cardanoLedgerSpecsPrsJSON": { "type": "githubpulls", "value": "input-output-hk cardano-ledger-specs", "emailresponsible": false }
          ,"walletPrsJSON": { "type": "githubpulls", "value": "input-output-hk cardano-wallet", "emailresponsible": false }
          ,"shellPrsJSON": { "type": "githubpulls", "value": "input-output-hk cardano-shell", "emailresponsible": false }
+         ,"cardanoNodePrsJSON": { "type": "githubpulls", "value": "input-output-hk cardano-node", "emailresponsible": false }
```
jbgi: should there also be `cardanoBasePrsJSON` in this set?

rvl: Yes there should, thanks, fixed.

PR #620, opened 2019-06-12
Fix `gac-new.sh`

kreisys:


kreisys:
[Fix `gac new`](https://trello.com/c/jLDaqGAZ/65-fix-gac-new)

PR #623, opened 2019-06-17
Adding monitoring for bithumb

neville-freeman:
Some new code for monitoring


disassembler:
At a bare minimum, we need to validate the process starts when we make a PR. In case you didn't know how to do this:

disassembler:
I pushed a commit fixing the indentation errors. Please do ensure your code works before creating a PR in the future @neville-freeman.

It still fails though after you run the process:
```
processing bithumb assets
failed to process bithumb assets
```

PR #639, opened 2019-07-16
cardano-node:  bump to serge/configuration-integration

deepfire:
Status: early preview of https://github.com/input-output-hk/cardano-node/pull/60

Adds two options to `cardano-node` -- `--genesis-file` and `--genesis-hash`:
  - https://github.com/input-output-hk/cardano-shell/blob/master/src/Cardano/Shell/Constants/CLI.hs#L44
  - https://github.com/input-output-hk/cardano-shell/blob/master/src/Cardano/Shell/Constants/CLI.hs#L49

# Status

The interface won't change much if at all, only caveat is potential bugs -- so I think it's safe to begin integration.

### DIFF
--- a/cardano-node-src.json
+++ b/cardano-node-src.json
@@ -1,6 +1,6 @@
 {
   "url": "https://github.com/input-output-hk/cardano-node",
-  "rev": "b9c17c39dcab94ae8a7e62d18616d1a4ad28ff22",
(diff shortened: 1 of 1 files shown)


disassembler:
Is this still supposed to be a draft or ready to merge?

deepfire:
I think it's uselessly outdated.

PR #642, opened 2019-07-26
Update ssh-keys.nix

craigem:
Add `craige`


craigem:
Resubmitting directly and with a pull signed by my IOHK key.

PR #643, opened 2019-07-26
Exchange monitoring

neville-freeman:
Bitmax monitoring


neville-freeman:
Bitmax alert

modules/exchange-monitor/exchange-monitor.py
```diff
@@ -120,3 +135,9 @@ def process_coinex_assets():
             print("failed to process coinex assets")
             coinex_active.set(False)         
         time.sleep(SLEEP_TIME)
```
disassembler: this sleep should be removed.

disassembler:
@neville-freeman this looks good. Please squash the commits into a single commit and force push and I'll approve/merge.

PR #656, opened 2019-08-10
Packet.net nixops deployments

disassembler:
Removes legacy packet.net machines and replaces them with packet.net nixops module.


deployments/packet-net.nix
```diff
@@ -2,14 +2,43 @@
...
+
     imports = [ ../modules/common.nix
                 module
-              ] ++
-              # This loads the packet.net-provided .nix files.
-              map (f: builtins.toPath(toString(./.) + "/${hostname}/" + f))
-                  (builtins.attrNames (builtins.readDir (./. + "/${hostname}")));
+              ] ++ (optionals (type == "legacy")
+              (map (f: builtins.toPath(toString(./.) + "/${hostname}/" + f))
+                  (builtins.attrNames (builtins.readDir (./. + "/${hostname}")))));
```
cleverca22: you probably dont want to use `toString` or `toPath`
just `./. + "/${hostname}/${f}` should do

johnalotoski: Updated

PR #670, opened 2019-09-04
niv for all sources

disassembler:
This switches to niv for all sources. Most of this is borrowed from cardano-node-wip. This should ease our pains of multiple repos being imported in multiple places across the code-base.


jobsets/cardano.nix
```diff
@@ -29,7 +28,7 @@ let
   };
 in pkgs.lib.fix (jobsets: {
   inherit (jobs) iohk-ops iohk-ops-integration-test nixops;
-  inherit (iohkpkgs) checks IFDPins cachecache;
```
cleverca22: the removal of `IFDPins` may make caching worse
it might be better to write a reusable function that can generate the pins from the niv `sources` attr

PR #702, opened 2019-11-19
buildkite-agent: provide docker credentials

rvl:


craigem:
Tested, Works as advertised.

bors r+

